PR #84, opened 2015-01-20
Future-proofing CasADi url

jgillis:


rgommers:
Thanks @jgillis, merged.

PR #97, opened 2015-07-03
Include imageio library in topical software

almarklein:


almarklein:
cc @kuchi

kuchi:
The link may be missing the 2nd ` character at the end.

almarklein:
Ah, thanks. Hooray for the weird syntax for links in rst :(

rgommers:
Looks good to me, merged.

rgommers:
The laptop I'm traveling with isn't set up properly to upload to the host server, so this change isn't immediately reflected on the site.

almarklein:
That's ok, as long as it gets there eventually :) thanks!

PR #119, opened 2015-10-24
News: Scipy 0.16.1 and Numpy 1.10.1 releases.

rgommers:


rgommers:
0.16.1 tag should be pushed tomorrow morning, when the binaries are built and uploaded.

Issue #102, opened 2015-08-10
Wiki and blogs are down (again)


DhrubajyotiDas:
Currently, neither http://wiki.scipy.org/ nor http://planet.scipy.org/ can be accessed.

Same issue as this one: https://github.com/scipy/scipy.org/issues/85

takluyver:
I've pinged people by email, but no response yet. I'll ping again.

argriffing:
"It's not just you! http://mail.scipy.org looks down from here."

filipesaraiva:
It is down yet :(

eternia478:
Mailing list is also acting up, with weird things like 

<code>
<scipy-dev-scipy.org@scipy.org> (expanded from <scipy-dev@scipy.org>): unknown user: "scipy-dev-scipy.org"
</code>

Not sure if related. Also, http://mail.scipy.org and http://planet.scipy.org are directed to MPI for Python, which doesn't make sense. wiki is completely down.

Naereen:
Wiki is still down, no-one has any idea how to fix that?

pv:
xref https://github.com/numpy/numpy/issues/6325

fkbreitl:
Same thing here. My bug reports were all closed.

imranal:
The http://wiki.scipy.org/ is still down. I can't access it.

rcoacci:
Is there any place where we can find the wiki content while it's down? Internet Archive has nothing, since the wiki had a restrictive robots.txt

Naereen:
I tried too, and Google cache, and other web crawler caches, *nothing* :cry: !

argriffing:
@rcoacci @Naereen Just out of curiosity, were you guys looking for the 'numpy for matlab users' page?

rcoacci:
Nope, but there are lots of really useful pages there like Cookbook and ParalellProgramming. I'm specifically after the second.

Naereen:
@argriffing yes!
I found [that one instead](http://mathesaurus.sourceforge.net/matlab-numpy.html), but it's somehow not as useful :disappointed: ...

argriffing:
Th mailing list has some [discussion](https://mail.scipy.org/pipermail/scipy-dev/2015-October/021012.html) about this, and at least some of the content is temporarily [rehosted](http://pav.iki.fi/tmp/to-be-removed/localhost_8080/NumPy_for_Matlab_Users.html).

imranal:
[...]

PR #121, opened 2015-10-25
Remove links to wiki.scipy.org

pv:
Let's pronounce wiki.scipy.org dead.

It's not been a wiki for many years (read-only), and it's been offline for the past month anyway. [There's also not very much material that's still useful in there](http://pav.iki.fi/tmp/to-be-removed/localhost_8080/AllPages.html). So best to not direct users there.

The Scipy-Cookbook perhaps could have a new life in a different form (cf https://github.com/pv/SciPy-CookBook/tree/master/ipython), but as it is now, much of the material is outdated and does things in an unnecessarily clumsy way. I removed links to it --- it may be more confusing than useful --- it's probably better to redirect people to read tutorials instead.

### DIFF
--- a/www/.htaccess
+++ b/www/.htaccess
@@ -1,14 +1,23 @@
 RewriteEngine on
 
 # Redirects to the old site, now at wiki.scipy.org
-RewriteRule ^(Cookbook|Developer_Zone|EuroScipy|F2Py|History_of_SciPy|License_Compatibility|MlabWrap|NumPy_for_Matlab_Users|Numpy_Functions_by_Category|PerformanceTips|PyLab|ReleaseNotes|Scipy200|SubmillisecondPulsar|Tentative_NumPy_Tutorial) \
-    http://wiki.scipy.org%{REQUEST_URI} [R=permanent]
+#RewriteRule ^(EuroScipy|History_of_SciPy|License_Compatibility|MlabWrap|PerformanceTips|PyLab|Scipy200|SubmillisecondPulsar) \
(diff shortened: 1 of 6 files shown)


rgommers:
+1

You may want to mention this on the mailing list. Then people can rescue stuff they still care about. And we can consider putting some redirects in for the most popular stuff.

pv:
The Additional_Documentation page btw also is fairly out of date,
http://pav.iki.fi/tmp/to-be-removed/localhost_8080/Additional_Documentation.html
With the exception of the "Astronomy tutorial" (available currently from the stsci/perry site in pdf form), it does not appear there's much of interest --- some of the stuff has been subsumed into numpy/scipy tutorials in any case.

pv:
I'll go forward on merging this --- at least as an interim solution.

argriffing:
Yes please merge this if it would settle the unrest at https://github.com/scipy/scipy.org/issues/102. I see that this PR doesn't just remove links as suggested by the title, but also adds and modifies some links.

pv:
It doesn't really solve the problem, because links to wiki.scipy.org are in the wild, also apparently in books. But at least best to not direct people knowingly to places that don't exist at the moment.

rkern:
If someone draws up a list of redirects, I'll make sure they get installed.

pv:
@rkern: 
A list of redirects at different location is at: https://github.com/scipy/scipy.org/blob/master/www/.htaccess
The rest of the pages probably can be redirected to http://scipy.github.io/old-wiki/

PR #157, opened 2016-03-26
Update installation options

### LINKED ISSUE #154: Please update the Debian version in the install page
The "Ubuntu & Debian" section of http://www.scipy.org/install.html page refers to Debian 7.0 which is an old version. The current stable Debian version is 8.3 and was released on Jan 23rd, 2016 [1]. Could you please update the page accordingly?

[1] - https://www.debian.org/releases/stable/

gfyoung:
Title is self-explanatory.  Closes #154.

### DIFF
--- a/www/install.rst
+++ b/www/install.rst
@@ -38,8 +38,8 @@ Ubuntu & Debian
 
     sudo apt-get install python-numpy python-scipy python-matplotlib ipython ipython-notebook python-pandas python-sympy python-nose
 
-The versions in Ubuntu 12.10 and Debian 7.0 meet the current SciPy stack
-specification. Users might also want to add the `NeuroDebian repository
+The versions in Ubuntu 12.10 or newer and Debian 7.0 or newer meet the current
+SciPy stack specification. Users might also want to add the `NeuroDebian repository
 <http://neuro.debian.net/>`_ for extra SciPy packages.
 
 Fedora
@@ -49,9 +49,7 @@ Fedora
 
     sudo yum install numpy scipy python-matplotlib ipython python-pandas sympy python-nose
 
-Users of Fedora 17 and earlier should then upgrade IPython using pip::
-
-    sudo pip install --upgrade ipython
+The versions in Fedora 17 or newer meet the current SciPy stack specification.
 
 Gentoo
 ~~~~~~
@@ -62,9 +60,7 @@ Gentoo
 
 You may get some messages saying that keyword changes or USE changes are
 necessary in order to proceed, and that you should use ``--autounmask-write`` to
-write changes to config files. This is especially likely to happen if you are
-running Gentoo Stable rather than Gentoo Testing, as of this writing (February
-2013).
+write changes to config files.
 
 If this happens, just run the above command with ``--autounmask-write``
 appended, then run ``sudo dispatch-conf`` (or an alternative) to save the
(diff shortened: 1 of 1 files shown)


www/install.rst
```diff
@@ -49,9 +49,9 @@ Fedora
 
     sudo yum install numpy scipy python-matplotlib ipython python-pandas sympy python-nose
 
-Users of Fedora 17 and earlier should then upgrade IPython using pip::
-
-    sudo pip install --upgrade ipython
+The versions in Fedora 23 meet the current SciPy stack specification. At time
+of writing (March 2016), Fedora 24 will soon be released (June 2016), so this
+section will be updated once testing can be done on that release.
```
ev-br: Maybe better avoid these sorts of promises, not sure how useful they are today.

www/install.rst
```diff
@@ -79,14 +77,32 @@ popular package managers you can install.
...
     sudo port install py27-numpy py27-scipy py27-matplotlib py27-ipython +notebook py27-pandas py27-sympy py27-nose
 
+Homebrew
+~~~~~~~~
+
+At the time of writing (March 2016), `Homebrew <http://brew.sh/>`_ does not
+have the full SciPy stack available (i.e. you cannot do ``brew install <formula>``
+for everything). However, it is a very popular option (perhaps even more so than
+Macports), and given that active work is being done in this `area <https://github.com/Homebrew/homebrew-python>`_,
+support for the entire SciPy stack should arrive soon. In the meantime, you can do
```
ev-br: This sounds like we're making a promise.

gfyoung: Perhaps I misworded it, as we're not the one promising anything.  This is work being done by the `Homebrew` group themselves.

www/install.rst
gfyoung: Fair enough.

www/install.rst
gfyoung: I changed the wording a bit to make it explicit that this is not work being done by `scipy` but rather by `Homebrew`.

www/install.rst
```diff
@@ -38,7 +38,7 @@ Ubuntu & Debian
 
     sudo apt-get install python-numpy python-scipy python-matplotlib ipython ipython-notebook python-pandas python-sympy python-nose
 
-The versions in Ubuntu 12.10 and Debian 7.0 meet the current SciPy stack
+The versions in Ubuntu 14.04 and Debian 8.3 meet the current SciPy stack
```
pv: Should say "versions newer than" probably rather than updating the numbers.

www/install.rst
```diff
@@ -79,14 +75,32 @@ popular package managers you can install.
...
 
+Homebrew
+~~~~~~~~
+
+At the time of writing (March 2016), `Homebrew <http://brew.sh/>`_ does not
+have the full SciPy stack available (i.e. you cannot do ``brew install <formula>``
+for everything). However, it is a very popular option (perhaps even more so than
+Macports), and given that active work is being done in this `area <https://github.com/Homebrew/homebrew-python>`_,
+by the `Homebrew group <https://github.com/Homebrew>`_ itself, support for the entire SciPy stack in the future
+seems likely. In the meantime, you can do ``brew install gfortran`` followed ``pip
```
pv: I'd remove everything here except the first sentence. This page shouldn't make predictions about the future. The `brew install gfortran` probably won't be enough eg for matplotlib.

www/install.rst
```diff
@@ -79,14 +75,32 @@ popular package managers you can install.
...
 
 .. _individual-packages:
 
+Windows packages
+----------------
+
+Windows does not have any package manager analogous to that in Linux, so installing
+one of the scientific Python distributions mentioned above is preferred. However, if
+that is not an option, Christoph Gohlke provides `pre-built Windows installers <http://www.lfd.uci.edu/~gohlke/pythonlibs/>`_
+for many Python packages, including all of the core SciPy stack, which work extremely well.
```
pv: I'd perhaps prefer to keep Gohlke's packages in the next section.

www/install.rst
```diff
@@ -79,14 +75,32 @@ popular package managers you can install.
...
 ~~~~~~~~
 
-To install the scipy stack for Python 2.7 with `Macports
+To install the SciPy stack for Python 2.7 with `Macports
 <http://www.macports.org>`_ execute this command in a terminal::
 
     sudo port install py27-numpy py27-scipy py27-matplotlib py27-ipython +notebook py27-pandas py27-sympy py27-nose
 
+Homebrew
+~~~~~~~~
```
pv: Maybe better to replace these 3rd level section headers by just boldface `**Homebrew**`.
This page should preferably be kept as short as possible, and the 3rd level headers start to make it look a bit messy.

www/install.rst
gfyoung: Well, they technically are Windows specific packages (or wheels to be precise) that you can install.  It only seems fair that if Linux and Mac get their own sections, Windows should get its own too?  Otherwise, I get the impression that these doc pages are more Linux-Mac oriented with Windows included on the side as somewhat of an after thought.

www/install.rst
gfyoung: That also works.  Done.

www/install.rst
gfyoung: Fair enough.  Done.

www/install.rst
gfyoung: Hmmm...boldface makes the section headers a little too small on my computer for my liking.  Their size starts to become indistinguishable from that of the section text.

gfyoung:
@pv : Do you have any updates with respect to my responses to your comments?  Otherwise, I think this is ready for merging.

PR #160, opened 2016-04-08
Add BLAS/LAPACK to Fedora installation. Fixes #159

### LINKED ISSUE #159: pip upgrade of scipy requires BLAS and LAPACK on Fedora/CentOS
https://www.scipy.org/install.html#fedora advises users to use `yum` to install scipy and then `pip` to upgrade it. However, upgrading via `pip` fails unless the user has BLAS and LAPACK devel packages installed.

Per http://stackoverflow.com/a/25147728/500207, the solution is to first
```
$ sudo yum install blas-devel lapack-devel
```
and only then `sudo pip install --upgrade scipy`.

fasiha:
Advises users to Fedora/CentOS/etc. to install dev versions of BLAS/LAPACK so that subsequent upgrades via `pip` will not fail.

### DIFF
--- a/www/citing.rst
+++ b/www/citing.rst
@@ -137,7 +137,7 @@ __ http://jmlr.org/papers/v12/pedregosa11a.html
 scikit-image
 ############
 
-* Stéfan van der Walt, Johannes L. Schönberger, Juan Nunez-Inglesias, François
+* Stéfan van der Walt, Johannes L. Schönberger, Juan Nunez-Iglesias, François
   Boulogne, Joshua D. Warner, Neil Yager, Emmanuelle Gouillart, Tony Yu and the
   scikit-image contributors.
   **scikit-image: Image processing in Python**,
--- a/www/index.rst
+++ b/www/index.rst
@@ -214,6 +214,9 @@ News
 .. role:: news-date
    :class: news-date
 
+SciPy 0.17.1 released :news-date:`2016-05-12`
+    See :doc:`/scipylib/download`.
+
 NumPy 1.11.0 released :news-date:`2016-03-27`
     See :doc:`/scipylib/download`.
 
--- a/www/index.rst
+++ b/www/index.rst
@@ -214,7 +214,7 @@ News
 .. role:: news-date
    :class: news-date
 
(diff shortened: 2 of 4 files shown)


charris:
Hmm, we are now putting linux binaries up on PiPI, so this may not be required. I'd also recommend atlas-devel instead of blas, and dnf instead of yum for recent fedora. Not sure where all that should go. @rgommers Thoughts?

rgommers:
Not a Fedora user, so no opinion on ``dnf`` vs ``yum. 

``atlas-devel`` (assuming it works) does sound preferable. @fasiha can you check if that works for you, and if so change to that?

takluyver:
`dnf` is the new `yum` on recent versions of Fedora (Fedora >= 23, I think). `yum` is aliased to it, so yum commands will still work for now. I'd guess it will remain for a decent length of time.

charris:
@rgommers Fedora moved to `dnf` in Fedora 22, so it is fairly new.

fasiha:
@rgommers sorry took a few days. `atlas-devel` and `lapack-devel` tested to work (atlas-devel has blas-devel as prerequisite). I tested on my ancient CentOS 6.5 VM, so I have no preference on `dnf` vs `yum`. Updated the branch and rebased to grab recent changes.

rgommers:
If ``dnf`` is new and ``yum`` still works, let's stay with ``yum`` for a little while I'd say - it works for a wider range of OS versions.

rgommers:
> Now that I look this big list of changes my PR is proposing… should I not have rebased my branch on master? Sorry for being open-source noob

No worries. Something did go wrong here, not sure how you managed to get those commits in here. You can either just add commits to an existing PR, or rebase on master and then force-push to the open PR. But it looks like you cherry-picked from master?

rgommers:
@fasiha this should work without ``lapack-devel``, ATLAS provides an optimized LAPACK. Can you confirm that that works for you?

rgommers:
Don't really want to spend the time fixing up this PR, since it's a one-liner. I cherry-picked the relevant commits, squashed them, removed ``lapack-devel`` and pushed the commit in f53c1ff01f.

PR #165, opened 2016-05-20
added noisyopt and scipydirect to python optimization topical software

andim:
Hi,

I would like to add two optimization packages to the topical software list.

The first, noisyopt https://github.com/andim/noisyopt, is a library for the optimization of noisy functions. It currently includes a stochastic perturbation stochastic approximation algorithm and a pattern search algorithm with adaptive sampling.

The second, scipydirect https://github.com/andim/scipydirect, is a wrapper around the DIRECT global optimization algorithm.

Both packages are lightweight and provide a scipy.optimize compatible call syntax to allow easy interchangeability with the optimization algorithms from scipy, which is crucial to allow fast testing of different algorithms on a given optimization problem.

Cheers
Andreas


ev-br:
Thanks Andreas.

PR #173, opened 2016-07-21
Try to be clearer that scipy.org bug tracker is not scipy bug tracker

pv:
It seems that the wording on the bug tracker page may be confusing.
Try to word it more clearly to indicate that the website bug tracker is not scipy bugtracker.


rgommers:
Makes sense, thanks Pauli,

PR #178, opened 2016-09-09
DOC: update install instructions with pip

matthew-brett:
Add pip to the install options.  Update homebrew instructions.

See: https://github.com/scipy/scipy/issues/6558

### DIFF
--- a/www/install.rst
+++ b/www/install.rst
@@ -28,12 +28,15 @@ include all the key packages:
 Installing via pip
 ------------------
 
-Mac and Linux users can install pre-built binary packages for the scipy stack
+Mac and Linux users can install pre-built binary packages for the SciPy stack
 using `pip <https://pip.pypa.io/en/stable>`_.  pip is the installation tool
-recommended by the Python Packaging Authority.
+recommended by the `Python Packaging Authority <https://www.pypa.io>`_. Pip
+can install pre-built binary packages in the `wheel
+<http://wheel.readthedocs.io>`_ package format.
 
-At the moment, *pip does not work well for Windows* because we do not have the
-necessary compilers to build some binary packages on Windows.
+*pip does not work well for Windows* because the standard pip package index
+site, `pypi <https://pypi.python.org/pypi>`_, does not yet have Windows wheels
+for some packages, such as SciPy.
 
 To install via pip on Mac or Linux, first upgrade pip to the latest version.
 For Python 3::
@@ -42,7 +45,7 @@ For Python 3::
 
 Replace ``python3`` with ``python2`` for Python 2 pip upgrade.
 
-Then install the scipy stack packages with pip.  We recommend a *user*
+Then install the SciPy stack packages with pip.  We recommend a *user*
 install, using the ``--user`` flag to pip.  This installs packages for your
 local user, and does not need extra permissions to write to the system
 directories::
--- a/www/install.rst
+++ b/www/install.rst
@@ -35,29 +35,37 @@ installation tool recommended by the Python Packaging Authority.
 At the moment, *pip does not work well for Windows* because we do not have the
 necessary compilers to build some binary packages on Windows.
 
-To install via pip on Mac or Linux, first upgrade pip to the latest version::
+To install via pip on Mac or Linux, first upgrade pip to the latest version.
+For Python 3::
 
-    python -m pip install --upgrade pip
+    python3 -m pip install --upgrade pip
+
+Replace ``python3`` with ``python2`` for Python 2 pip upgrade.
 
 Then install the scipy stack packages with pip.  We recommend a *user*
 install, using the ``--user`` flag to pip.  This installs packages for your
 local user, and does not need extra permissions to write to the system
 directories::
 
-    pip install --user numpy scipy matplotlib ipython jupyter pandas sympy nose
+    pip3 install --user numpy scipy matplotlib ipython jupyter pandas sympy nose
+
+Use ``pip2`` instead of ``pip3`` for Python 2 installs.
 
 For user installs, make sure your user install executable directory is on your
-PATH.
+PATH.  Here are example commands for setting the user PATH where the user is
+using Python 3 by default:
 
 Linux::
 
+    # Consider adding this at the end of your ~/.bashrc file
     export PATH="$PATH:/home/your_user/.local/bin"
 
 OSX::
 
+    # Consider adding this at the end of your ~/.bash_profile file
     export PATH="$PATH:/Users/your_user/Library/Python/3.5/bin"
 
-replacing ``your_user`` with your username.
+Replace ``your_user`` with your username.
 
 Install systemwide via a Linux package manager
 ----------------------------------------------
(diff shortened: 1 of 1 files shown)


www/install.rst
```diff
@@ -26,10 +25,47 @@ which includes all the key packages:
...
+PATH.
+
+Linux::
 
-Users on Linux can quickly install the necessary packages from repositories.
+    export PATH="$PATH:/home/your_user/.local/bin"
+
+OSX::
+
+    export PATH="$PATH:/Users/your_user/Library/Python/3.5/bin"
```
toastedcornflakes: I think that `python -m pip` will use python2/pip2. The executables would then end up in `…/Python/2.7/bin` ?

matthew-brett: Yes, good point.

toastedcornflakes:
Looks good to me, thanks a lot 😃

www/install.rst
```diff
@@ -6,13 +6,12 @@ These are instructions for installing :ref:`the full SciPy stack
...
 SciPy, see :ref:`individual-packages` below.
 
-
 Scientific Python distributions
 -------------------------------
 
-For most users, especially on Windows and Mac, the easiest way to install the
-packages of the SciPy stack is to download one of these Python distributions,
-which includes all the key packages:
+For most users, especially on Windows, the easiest way to install the packages
```
rgommers: I would leave ``and OS X``. A full distribution is still strongly preferred over ``pip`` or downloading a package manager, less steps and better quality control.

www/install.rst
```diff
@@ -26,10 +25,55 @@ which includes all the key packages:
...
+necessary compilers to build some binary packages on Windows.
+
+To install via pip on Mac or Linux, first upgrade pip to the latest version.
+For Python 3::
+
+    python3 -m pip install --upgrade pip
+
+Replace ``python3`` with ``python2`` for Python 2 pip upgrade.
+
+Then install the scipy stack packages with pip.  We recommend a *user*
```
rgommers: A couple of people just went through the NumPy code base and capitalized everything, so I feel obliged to say ``SciPy Stack`` please:)

www/install.rst
```diff
@@ -26,10 +25,55 @@ which includes all the key packages:
...
+install, using the ``--user`` flag to pip.  This installs packages for your
```
rgommers: maybe add ``(unless you're inside a virtualenv)``. because then ``-user`` is not what you want

www/install.rst
```diff
@@ -26,10 +25,55 @@ which includes all the key packages:
...
---------------
+Installing via pip
+------------------
+
+Mac and Linux users can install pre-built binary packages for the scipy stack
+using `pip <https://pip.pypa.io/en/stable>`_.  pip is the installation tool
+recommended by the Python Packaging Authority.
+
+At the moment, *pip does not work well for Windows* because we do not have the
+necessary compilers to build some binary packages on Windows.
```
pv: We don't need to give an explanation why Windows is not available currently.

www/install.rst
matthew-brett: Is the quality control for the scipy stack better on Anaconda etc than for the pip packages?

rgommers: I think so, because the exact versions that users install are tested together. And everything is built against the exact numpy version that is used by the user. With ``pip`` it's a more random mix.

The more important reason is less moving parts though: just download one thing and it all works. Plus you get MKL with Anaconda, which also is nice.

matthew-brett: As far as I know, we have no outstanding bugs for numpy or scipy wheels on Linux or OSX.  We tested the hell out of the numpy and scipy wheels via : https://github.com/matthew-brett/manylinux-testing and http://build.openblas.net .  The wheels are commonly used via travis-ci integration testing.    As far as I know the wheels (using OpenBLAS and Accelerate) are in the same ballpark as MKL packages for speed (not so for Windows of course).   We (the numpy and scipy developers) do support the wheels, but do not directly support the Anaconda / Canopy etc packages.

I think it's reasonable to recommend a big distribution here, as we do in the current text ("For most users ... the easiest way to install"), and it's really the only way to go for Windows, but I think it's too strong to say "especially for OSX".

rgommers: I agree that the situation on OS X is hugely better than on Windows. I just want to ensure that it's very clear to every new user that comes in, he/she should use a distribution. Actually, I should correct myself: Anaconda or Miniconda. It is the best choice, for many reasons. The text ``For most users, especially on Windows ...`` doesn't really do that well enough.

There's an issue with us recommending one over the others, but on the other hand we do new users a big disservice if we don't point them to the best option. 

With ``conda`` and ``conda-forge`` we now have things that are fully open source and community-oriented. Oriented on *our* community, unlike ``pip`` and ``PyPI``. I see the differences in usability and capabilities only getting bigger, looking at how fast ``conda`` and ``conda-forge`` move and how disfunctional decision making for ``pip`` is.

> We tested the hell out of the numpy and scipy wheels 

I know, and your efforts are much appreciated! The situation has improved massively. 

> do not directly support the Anaconda / Canopy etc packages

Actually, we do for conda-forge. I'm one of the maintainers for ``numpy`` and ``scipy`` now, and many other contributors and core devs are involved for those and other packages.

rgommers: How about we merge this PR as is, and I can try to write something to better guide the new user?

rgommers: Note, the above doesn't mean I stop working on pip/wheels or participating in distutils-sig. The latter activity is pretty frustrating though.

matthew-brett: We can do the merge, but I think we disagree on the how much better conda + infrastructure is than pip + infrastructure, so I guess we'll end up having the same discussion on the new PR.

Don't get me wrong, I'm happy to broadly recommend conda etc even though I do not use or recommend them myself.   My concern here is to remain relatively even-handed in the situation where there are trade-offs for each.

rgommers: We should definitely explain the trade-offs. If you have an existing Python install for example, or you don't just do scientific work but also web/desktop/etc development, ``pip`` may be a better fit. The problem is that we don't explain that now anywhere. The result is that way too many new users choosing things like python.org installers, and once they learn more they vote with their feet - towards a distribution (and ``conda``), hardly ever the other way around.

matthew-brett: Maybe put a document on the trade-offs elsewhere and link to it from this document?

I personally recommend homebrew and pip on OSX unless you have very specialized Python requirements: https://bic-berkeley.github.io/psych-214-fall-2016/installation_on_mac.html

But - I was deliberately trying to avoid getting into that discussion in this PR.

rgommers: > Maybe put a document on the trade-offs elsewhere and link to it from this document?

good idea, will do

> But - I was deliberately trying to avoid getting into that discussion in this PR.

that didn't quite work out, sorry:)

rgommers:
Merged, thanks Matthew.

rgommers:
Will push these to scipy.org later, not on my development machine now. Or someone else can do it now.

www/install.rst
rgommers: Follow-up in PR gh-179

PR #179, opened 2016-09-10
ENH: add page to discuss how to choose an installation method.

rgommers:


pv:
I'm somewhat worried that these pages are getting too long, and appear confusing and complicated for newcomers.

I think the point of the install page to point the method/recipe that we recommend for each platform, but not necessarily to list or discuss all possible methods.

www/choosing-install-method.rst
```diff
@@ -0,0 +1,149 @@
...
+- Can work with an existing Python install (do *not* install into the Python
+  provided by your OS though, that's always a bad idea).
+- Is very widely used, and is the installer recommended by the
+  `Python Packaging Authority <https://www.pypa.io>`_.
+- Installs wheels, which more and more packages provide on PyPI_.
+
+Cons:
+
+- Does not work well on Windows.
+- Does not include a true dependency resolver (so state can become inconsistent).
```
matthew-brett: I've never encountered this.  I can't think offhand think of that problem coming up in an issue or mailing list question.

www/choosing-install-method.rst
```diff
@@ -0,0 +1,149 @@
...
+- Cannot handle non-Python dependencies (those are important for scientific
```
matthew-brett: Well, there are wheels which use BLAS / HDF5 etc, with the libraries shipped in the wheel. so I think this one is more of a technical note.

www/choosing-install-method.rst
```diff
@@ -0,0 +1,149 @@
...
+  applications: BLAS/LAPACK, HDF5, etc.).
+- Not all packages provide complete dependency info on PyPI, which can make
+  pip_ installs fail.
+- Publishing wheels on PyPI (so pip_ can install them) is hard, especially
```
matthew-brett: This also seems like a debatable technical point.

matthew-brett:
I also think this page is too long, and reads more like a pitch on "why I like conda".

www/choosing-install-method.rst
```diff
@@ -0,0 +1,149 @@
...
+  `Python Packaging Authority <https://www.pypa.io>`_.
+- Installs wheels, which more and more packages provide on PyPI_.
+
+Cons:
+
+- Does not work well on Windows.
+- Does not include a true dependency resolver (so state can become inconsistent).
+- Cannot handle non-Python dependencies (those are important for scientific
+  applications: BLAS/LAPACK, HDF5, etc.).
+- Not all packages provide complete dependency info on PyPI, which can make
```
matthew-brett: You are maybe thinking of packages that don't declare dependencies on numpy / scipy etc.   This can happen, but a) is not very common now b) doesn't apply to the instructions on the installation page and c) is easily fixed.

Do you know of any current examples?

www/choosing-install-method.rst
```diff
@@ -0,0 +1,149 @@
...
+Pros:
+
+- Package manager itself already installed (on Linux) or so useful for
+  non-Python things that you'll want to install it anyway (on OS X).
+- Package managers are reliable, mature technology.
+- Easy to get started.
+
+Cons:
+
+- On Linux, installs into the same Python environment as used by the OS itself,
```
matthew-brett: I believe this is a) a rare problem b) solved by user installs, which we recommend.

www/choosing-install-method.rst
```diff
@@ -0,0 +1,149 @@
...
+  which can be fragile if you want to install packages also with pip_.
+- On OS X, packages provided by Homebrew and Macports are often not as well
```
matthew-brett: I guess you must be referring to BLAS / LAPACK?   A quick test suggests that homebrew uses Accelerate for numpy BLAS, at least.  What reason do you have to think that homebrew or macports packages are not optimized and / or tested properly?

rgommers:
> I'm somewhat worried that these pages are getting too long, and appear confusing and complicated for newcomers.

That's why I split it into 2 pages, one "how do I install using methods X, Y, Z" for those who know what they want (with a single recommendation to use a distribution), and the other one with how to choose a method. You wouldn't get to this page at all from the main scipy.org page. Only if youre' on scipy.org/install and really don't know, then follow the link. 

I'm happy to link to an external reference for this, but I'm not aware of a good one that is aimed at scientific users.

> I think the point of the install page to point the method/recipe that we recommend for each platform, but not necessarily to list or discuss all possible methods.

Hmm, that would make it a very short page. If I take that literally, then we'd have to remove 80% of content. I'd rather aim for the commands/downloads for all popular methods. Otherwise I'd have to strongly advocate for removing pip from the whole page (which seems wrong), because for new users it's not the best method anywhere.

www/choosing-install-method.rst
rgommers: Eh, not really. It's the same issue as why we say on OS X to never use system Python.

www/choosing-install-method.rst
rgommers: This I took from your PR. You just added some text saying to not use Homebrew provided packages even if they exist but use ``pip`` instead. So you tell me:)

www/choosing-install-method.rst
rgommers: This is the root cause of many ``pip`` issues. I've broken packages by upgrading other ones. But here's an example of an issue with a ridiculous amount of +1's that's caused by this: https://github.com/pypa/pip/issues/59

This is probably the single worst problem pip has.

www/choosing-install-method.rst
rgommers: Also no, that's not just a technical note. Not long ago I wanted PyQt on Linux for Python 3.x. There were no 3.x package manager packages. With ``pip`` this is basically not possible. I gave up after wasting 2 hours.

And yes, technically a wheel is a zip so you can stuff anything inside it. But almost no one does that because it's ridiculously hard.

www/choosing-install-method.rst
rgommers: SciPy itself for example will not correctly declare the lowest numpy version it supports. See a long post by Nathaniel on distutils-sig for details: https://mail.python.org/pipermail/distutils-sig/2015-October/027161.html

www/choosing-install-method.rst
rgommers: I can tweak the text, but it's extremely clear that it's far harder than publishing conda packages. It requires a lot of nontrivial steps. Last year I did it for PyWavelets (only C, no BLAS etc.), probably took a day to get Appveyor to work well, then you have to download things from there, etc. Still don't have OS X wheels. There are just very many moving parts, and little automation.

rgommers:
> I also think this page is too long, and reads more like a pitch on "why I like conda". 

We can remove the pros/cons if you all prefer that, I'll find another home for them. What about the recommendations?

rgommers:
> reads more like a pitch on "why I like conda". 

I'm sorry, but that's not really a fair comment. The technical differences in capability have grown so large that it's not about personal preference (and I personally use both). ``pip`` does have its place, it's just not the first thing we should be teaching new scientific users.

matthew-brett:
I'll respond to the individual things if you like, but how about a thought experiment.  Why don't I write a page saying "why I recommend pip" (which I do) and you write one on "why I recommend conda" (which you do).

I say thought experiment, because that doesn't seem like something we should do on the scipy.org website, but on blog pages or whatever.

I think we should stick with the minimum stuff we can all agree on, and I think that's covered by the original install page.

rgommers:
> "why I recommend conda" (which you do).

Actually no, for new users I recommend Anaconda first, and any other scientific Python distribution second. Well before either ``pip`` or ``conda``. That to me is much more important than the ``pip`` or ``conda`` thing, which users can start worrying about once they've become productive and happy users.

One other thought: maybe we should add one or two sentences about Jupyter notebooks / Spyder / IPython. I see a large amount of new users trying IDLE or a plain Python shell, which of course gets them nowhere fast.

> I think we should stick with the minimum stuff we can all agree on, and I think that's covered by the original install page.

To be honest, looking at that as is, can't quite agree with the current version. We really should not be talking to new users about things like messing with ``.bash_profile``. So either we make it clear that that's not for them (like I'm trying by recommendation for new users and other types of users), or we remove the content.

> I say thought experiment, because that doesn't seem like something we should do on the scipy.org website, but on blog pages or whatever.

I'm happy to do that, although I probably won't have too much to add on top of what I wrote and @jakevdp 's excellent blog post.

So how about we do that, and continue the discussion here only on "what should be on scipy.org/install?" for now. Is it:

- Instructions for only the preferred install method per platform?
- Instructions for all commonly used install methods?
- Recommendations for new users?
- Recommendations for several types of users?

www/choosing-install-method.rst
matthew-brett: See : https://travis-ci.org/MacPython/pywavelets-wheels  .  I just hacked that up in the last 25 minutes with no previous experience of the PyWavelets codebase.  It's currently building and testing and uploading OSX and manylinux 32- and 64-bit wheels for Pythons 2.7, 3.4 and 3.5.   Have a look at the build repo for the amount of configuration I needed to do :  https://github.com/MacPython/pywavelets-wheels .

takluyver:
I'm on my phone and haven't read this stuff properly, but I'd echo the
concern that this feels like exposing new users to a lot of complex
information that will be more off-putting than useful to them. I worry that
people are more interested in promoting their preferred packaging system,
and the efforts to find a compromise will end up pushing the decision onto
people who can't easily make it.

For Jupyter, we (try to) have a very simple main install page: we recommend
Anaconda for users new to Python, give brief instructions for that, and
also give instructions using pip for users who already have a Python set up
which they're comfortable with. This isn't neutral, but we think new users
are better served by a simple recommendation than a list of pros and cons.

www/choosing-install-method.rst
matthew-brett: Wheels now all built, tested and uploaded to Rackspace container.   Available at https://3f23b170c54c2533c070-1c8a9b3114517dc5fe17b7c3f8c63a43.ssl.cf2.rackcdn.com or http://wheels.scipy.org/  . Pypi upload instructions in https://github.com/MacPython/pywavelets-wheels/blob/master/README.rst are:

```
VERSION=0.4.0
CDN_URL=https://3f23b170c54c2533c070-1c8a9b3114517dc5fe17b7c3f8c63a43.ssl.cf2.rackcdn.com
wheel-uploader -r warehouse -u $CDN_URL -s -v -w ~/wheelhouse -t macosx PyWavelets $VERSION
wheel-uploader -r warehouse -u $CDN_URL -s -v -w ~/wheelhouse -t manylinux1 PyWavelets $VERSION
```

rgommers:
Thanks @takluyver. I think http://jupyter.readthedocs.io/en/latest/install.html looks excellent, I'd be in favor of modeling our main install page on that and put all other content elsewhere.

www/choosing-install-method.rst
rgommers: Thanks Matthew for proving me wrong on the OS X part, and producing something quite useful doing so:)

It's still a bit more work than for ``conda-forge`` (with the encrypted keys, download/upload), but not a whole lot.

matthew-brett:
I consider Jupyter's installation page at being the extreme end of Anacondification.  Compare http://matplotlib.org/users/installing.html for a more balanced example.

www/choosing-install-method.rst
matthew-brett: Another 10 minutes seems to have solved the Windows build as well : https://ci.appveyor.com/project/matthew-brett/pywavelets-wheels

www/choosing-install-method.rst
matthew-brett: I agree that that 'upgrade all' behavior is annoying.  I haven't myself broken anything with it, despite using it almost every day for several years, but it's certainly frustrating, to the point of being a bug.    As you know, this should be resolved relatively soon by a change in the default behavior or a new preferred command.   At that point pip will still lack a full dependency resolver, but it will be difficult to notice that lack.

www/choosing-install-method.rst
matthew-brett: The current standard way of solving that is to ship the library in the wheel.   It's pretty easy to do that automatically with `delocate` and `audit-wheel`.   I suppose one might prefer separate packages, but that does seem to me to be a technical note.

www/choosing-install-method.rst
matthew-brett: Right.  For people who have not read Nathaniel's post, the problem was that, until recently, wheels have not been available for numpy, and, in combination with the 'upgrade all' behavior of pip, it was common for a statsmodels or scipy pip install to generate a numpy install from source.    I see in the Scipy setup.py that there's a check for the 'darwin' platform just for that reason.  `pip install scipy` on OSX does install numpy.    The current (very old) version of statsmodels doesn't do that, but it should.

matthew-brett: * https://github.com/statsmodels/statsmodels/pull/3195
* https://github.com/scipy/scipy/pull/656

takluyver:
Matplotlib is certainly more balanced, but 'balance' is not our primary
goal: we want to make it easy for new users to get started. I don't like
how even the first part of the mpl instructions recommends anaconda or
miniconda or canopy. Should the user just pick the name they like best? One
of the three does not even include mpl out of the box, which it totally
fails to explain.

This is especially important for Jupyter, because more people want to
install it who don't already know about Python. But in general I think
simplicity should trump balance in our most visible installation
instructions.

matthew-brett:
I do think the Jupyter install instructions are rather extreme.    You start with "We recommend using the Anaconda distribution to install Python and Jupyter.".  Then you have the heading "Installing Jupyter using Anaconda and conda", followed by "For new users, we highly recommend installing Anaconda. "  Under the next large heading "Optional for experienced Python developers: Installing Jupyter with pip" you have the first sentence "We recommend Anaconda for installing Jupyter. ".  This level of emphasis might be reasonable if it was difficult to install Jupyter with pip, but it's extremely easy.  So  I think these would qualify as activist install instructions.  

For comparison, you've already seen the matplotlib instructions (which could no doubt be improved).  The scikit-image instructions mention conda etc second after pip : http://scikit-image.org/docs/stable/install.html .   Same for scikit-learn : http://scikit-learn.org/stable/install.html .

It's totally reasonable to recommend Anaconda for a new user.   I wouldn't recommend it myself, but I completely understand why someone might do that.  That is what we already do in our install page.

On the other hand, there are many people who would prefer other install methods, for various good reasons, and it seems to me perfectly reasonable to list the alternatives, for those with the energy to read a bit further down the page.

www/choosing-install-method.rst
matthew-brett: System Python on OSX promotes its own numpy / scipy /matplotlib above any version that you install.  There's also the vague worry that you might damage the system install of Python, but I haven't succeeded in doing that myself, on Linux or OSX.

www/choosing-install-method.rst
matthew-brett: My reasoning for recommending pip over homebrew was that the pip wheels cover more packages, so it's easier to stick to one method rather than mixing homebrew and pip installs.   I didn't mean to suggest the homebrew installs were poorly tested or optimized, I don't have any reason to think that.

takluyver:
We may be repeating that recommendation more than necessary; I'll look at
tuning it down a bit. But I really don't think we're being Anaconda
activists. We tried to consider different situations in which people might
read the install page, and what we'd recommend if such people asked us in
person how to install it. That led to the realisation that two
recommendations cover ~95% of what we'd tell users: if you already have
python and know what you're doing, install with pip. Otherwise, get
Anaconda.

www/choosing-install-method.rst
rgommers: OK, then we should not mention that at all I think.

www/choosing-install-method.rst
rgommers: > As you know, this should be resolved relatively soon by a change in the default behavior or a new preferred command. 

I wish that were true, but I'm not optimistic. There's agreement on making the change, but not on how. There are 2 good PRs and a PEP-style write-up, but the pip/PyPA decision making on this is so incredibly disfunctional that it may not happen for another couple of years.

www/choosing-install-method.rst
rgommers: Well, that's all pretty hard to do for the first time and also not very discoverable for your average package author.

You know all this stuff really well, but the amount of packages with compiled code that have complete wheels (or even Windows + OS X) up on PyPI which you had no hand in is very small.

rgommers:
> We tried to consider different situations in which people might read the install page, and what we'd recommend if such people asked us in person how to install it. 

That is a good argument. It's very politically correct to treat everything equally, but doesn't really help users. I don't think many of us will have ever recommended Pyzo or Canopy in person as the best option to use, so why do it here?

Scikit-image even mentions Bento. As much as I like it, that's not for people who need install instructions.

Scikit-image has in big bold letters *We don’t recommend installing scipy or numpy using pip on linux*. Avoid it on Windows as well. So that's one platform left.....

takluyver:
Maybe someone should make a separate site with a directory of all the options for installing scientific Python machinery, with pros and cons for each. ;-) I'm sure that's interesting to someone (maybe people like us trying to design install docs), but new users trying to get started are not that someone.

Eric89GXL:
In my experience (Ana)conda has made things easier in terms of getting users going. In libraries I work on we put in a Jupyter-like "use Anaconda" recommendation in the last year, and our installation-related issues seem to have decreased e.g. when teaching people how to use Python and our software in workshops.

How about we keep the page short (ideally 2-3 of screens worth of text/tables):

1. First top-level heading: Installation using Anaconda. We can add a sentence early on with a `.. note::` like "... note:: Anaconda is not a unanimously recommended -- some users and developers may prefer pip because it is <plug for pip>, or other methods. See < second-level heading link> for more details.`
2. Second top-level heading: Installation with other methods. It can even have a little table with pros and cons. We can order them roughly according to popularity (`pip` first at least). This doesn't have to be complete instructions for each. Complicated / long methods should linked to separate docs as necessary.
3. Third top-level heading: advanced/developer instructions, if needed. Preferably again brief, linking to another document if necessary.

@matthew-brett are you okay with that? I know it's still Anaconda-biased / Anaconda-promoting, but I do think we should promote one option first to simplify things for users, and Anaconda seems to have plurality support.

BTW @matthew-brett do the pip wheels allow use of Intel MKL? (Ana)conda makes it pretty easy, which is nice in terms of getting speedups for free.

matthew-brett:
Eric - the situation with MKL is the same as for Fortran compilation on Windows.  Because Continuum is a company, it can afford to distribute software binaries that have conditions unacceptable to community-based open-source projects.   See: 

* https://mingwpy.github.io/blas_lapack.html#intel-math-kernel-library
* https://software.intel.com/en-us/comment/1881526

That said, the Linux and OSX wheels use OpenBLAS and Accelerate respectively.  I suspect that these are both competitive with MKL for speed.

matthew-brett:
For the scikit-image install instructions - yes they are slightly out of date, in that they don't allow for the recent introductions of Linux wheels.   @stefanv - maybe you could comment here.

takluyver:
> We can add a sentence early on with a .. note:: like "... note:: Anaconda is not a unanimously recommended -- some users and developers may prefer pip...
>
> Second top-level heading: Installation with other methods. It can even have a little table with pros and cons.

This still seems like unnecessary confusion for new users. Telling them about our disagreements does not help them get the SciPy stack installed.

While I think Anaconda is generally the easier option for new users, my point in this thread is that we should give clear, simple instructions that require as few decisions as possible from people wanting to use our software.  Making tables of pros and cons is sacrificing the user experience to avoid developers having to reach a decision.

jakevdp:
I second @Eric89GXL's suggestion, though I'd even leave-out the "note" on our disagreements here. Our goal should be streamlining the experience for users, not assuaging internal tool debates 😄

matthew-brett:
I don't think 'politically correct' is the right term, but 'political' might be.

There are several deep issues going on here, about the future of the scientific Python stack, who owns it, how it will be maintained, and the relationship to the broader Python community.

I agree with "Getting going with Anaconda is the easiest way".    Whether it's the best way ahead for our community is a different and, in my view, much more important question.

I guess I was trying to avoid the rest of that question because it gets people very upset.  For example, the last time I painted a picture where Anaconda was the de-facto standard for scientific Python, the experienced Python developer that I was talking to said "OK, I'll switch to Julia".

By "balance", I mean, that rather than simply consider which is the easiest install option, we should bear in mind that we have a range of users with a range of ability, patience and (very strong) preferences.

The reason I like the current doc is that it stays neutral on that.  It doesn't raise these issues, recommends Anaconda as easy (for those whose primary desire is "easy"), but gives good air time to other methods that people like me who think that that is the wrong way to go.   For those whose only concern is easy, they only have to read the first paragraph, and the fact that the page continues is not going to upset them.

jakevdp:
@matthew-brett so your overall proposal is to close the PR and leave things unchanged?

takluyver:
I focus on ease because I figure that experienced users won't be relying on the install page anyway. Maybe that's not as true for things like Scipy with complex dependencies, but I'd still lean towards having a short 'install' page for new users and a separate page called something like 'advanced installation options'.

matthew-brett:
For this particular PR - yes, my preference would be to close it on the basis that these notes should stay on territory where there is consensus, and that we should in general take no greater stand than the recommendation we have now (before this PR).

Otherwise, I think we're going to have to elaborate a big picture about the future of scientific computing in Python, which is surely worth doing, but I don't think we should do it here.

matthew-brett:
Was just talking to @NelleV about this discussion at lunch today, who thought that @GaelVaroquaux would also be interested.

rgommers:
All, FYI the latest changes that were merged haven't yet been pushed to the site, here it is: https://github.com/scipy/scipy.org/blob/master/www/install.rst

> we have a range of users with a range of ability, patience and (very strong) preferences.

Yes, but the people at the high end of the ability range and with very strong preferences are **not** the audience of this page. 

> I agree with "Getting going with Anaconda is the easiest way". 

If we all agree on that, then that should be made as clear as it can be. I'm absolutely not interested in playing politics on the main install page of the whole stack at the expense of new users.

> de-facto standard for scientific Python, the experienced Python developer that I was talking to said "OK, I'll switch to Julia".

That seems so ridiculous it's funny. Unless he/she thought we were actively dropping support for other installation methods, which we don't have any plans for (the opposite actually, we're improving things on all fronts).

rgommers:
I'll put another PR which just aims at cleaning up the current mess (like outdated Gentoo instructions etc.). Maybe we can get to something more user-friendly in smaller steps.

pv:
As said above, I would suggest to keep the install page simple and very
aimed at beginners. Simple instructions that are likely to work. It can be
said at that other ways also exist and say eg. that pip  works. But I think
the page should be kept short.
If you look at stackoverflow or our bug trackers, the installation problem
questions are mainly on windows, and mostly about having installed some
unknown binary packages, installing gohlkes packages wrong, copying stuff
manually to sitepackages etc., also some cases of pip failing to download
binary wheels and falling back to compilation. Not very many anaconda etc
problems, or problems with linux system packages.

pv:
By the way, does double clicking wheels on windows open them in explorer as
zip files? If yes, someone at pypa probably should think about changing
that...

matthew-brett:
My point is that heavily promoting Anaconda at the expense of the other methods is "playing politics".   Leaving the page as it is reasonably neutral, and not playing politics.

I could argue that installing MATLAB was easier than installing all the packages and GUIs that i need for Python, and recommend that, but ease of installation is not my only criterion.

It's hard for me to understand why the current page is not adequate for people whose only aim is to get going as quickly as possible.   Why would such a person keep reading if they have the solution they want at the top of the page?

Remember that pip packages are relatively new.  So some older install instructions do not take them into account, and older stackoverflow questions will likewise be less relevant.

It seems to me we often have Anaconda package installation issues, at least I have found myself pointing people to the Anaconda issues page several times.

pv:
My comments of course apply to the main install page, I didnt read this PR.
Not a big fan of pro/con lists however, and I think the scipy.org landing
page is probably not a good place for extended discussion about merits of
each.

13.9.2016 1.39 "Matthew Brett" <notifications@github.com> kirjoitti:

matthew-brett:
> That seems so ridiculous it's funny.

I remember when I was in practice obliged to install Perl via an ActiveState installer.  I looked miserably at the icon for a while before I clicked it, and it was a great relief when I didn't have to do that any more.

The question in open source is always - who owns this?  Whose is it?   Who gets to say what happens next?  Who are "we" and where is our community?  I think it would be foolish to ignore those questions.

www/choosing-install-method.rst
```diff
@@ -0,0 +1,149 @@
...
+Cons:
+
+- Cannot work with an existing Python install.
+- Cannot use wheels (use conda binary packages instead), so for projects
+  not yet packaged for conda_ one still needs to use pip_ (or do the
+  packaging, which is fortunately easy).
+- Documentation and usage instructions are often sparser than for pip_.
+- If you're on a less commonly used OS (like 32-bit Linux), there will be
+  no precompiled packages.
+
```
NelleV: I'd also add that users have to trust the source of the pre-compiled binaries (by a company). That may not be a concern for most users, but that's a huge security concern for others. Some feel that's a reason big enough not to use conda, which isn't reflected here.

www/install.rst
```diff
@@ -118,7 +136,7 @@ appended, then run ``sudo dispatch-conf`` (or an alternative) to save the
 configuration changes, and finally run the original command again.
 
 Install systemwide via a Mac package manager
```
NelleV: systemwide -> system-wide?

NelleV:
My 2 cents.

*  I really don't like pros and cons. In addition, they are often biased.
* Conda and Anaconda have their shares of problems, that are not really reflected in the pros and cons. Out of the problems I've had recently with anaconda/conda, here are a couple:
    * They modify the bashrc. That recently messed up a friend's bashrc quite badly. None of her company's tool worked anymore, and it took a while to fix the damage (I don't know exactly what happened). Note that it also means that they won't work out of the box for people who uses other bashrc. That wasn't a great introduction to python for her.
   *  I ran into a bunch of problems recently using conda. For example, a hand-built version of matplotlib segfaults on my machine with the numpy version of conda. No clue why and impossible to debug. On the other hand, it works perfectly well with an environment installed with pip. I didn't have time to investigate more what was going on with the numpy version coming from conda.
* The main page looks fine like this. I'd just move anaconda *after* the package manager system. IMO, package manager systems are more reliable than anaconda and should always be the go-to way to install software.

takluyver:
> The main page looks fine like this. I'd just move anaconda after the
package manager system. IMO, package manager systems are more reliable than
anaconda and should always be the go-to way to install software.

If you mean system package managers like apt, I definitely disagree. That
involves a list of different instructions for each platform, including
different Linux distros. It doesn't cover Windows at all. And users on
major Linux distros will often get severely outdated versions like that.

For a sysadmin setting up a server, system packages may be a reliable,
trusted option. But I wouldn't recommend new users start like that.

rgommers:
The pros/cons were not on the main page and I've already said I would remove them. To avoid more arguments against something we're agreed on, I'll close this PR.

rgommers:
> I'd also add that users have to trust the source of the pre-compiled binaries (by a company). That may not be a concern for most users, but that's a huge security concern for others. Some feel that's a reason big enough not to use conda, which isn't reflected here.

That's (a) about a very small fraction of users, and (b) applies to other distributions and binaries as well so has nothing in particular to do with conda. And the few people who care about that already know what they like and don't like, so not the audience of this page.

rgommers:
> I remember when I was in practice obliged to install Perl via an ActiveState installer. I looked miserably at the icon for a while before I clicked it, and it was a great relief when I didn't have to do that any more.

No one is obliged to use any particular install method, so incorrect analogy.

> The question in open source is always - who owns this? Whose is it? Who gets to say what happens next? Who are "we" and where is our community? I think it would be foolish to ignore those questions.

An extended blog post may be helpful here, because this does feel like scaremongering and a few undeserved digs at Continuum. At least I've seen zero evidence of bad intentions on their part, quite the opposite actually.

rgommers:
basic cleanups without changing recommendations or structure of the page at gh-182

matthew-brett:
> No one is obliged to use any particular install method, so incorrect analogy.

Note that I said "I was in practice obliged to install".   Here we are thinking about what the future would look like if Anaconda became the only practical way to install the scipy stack.   I don't consider that an unreasonable outcome if we push our users aggressively towards Anaconda.  Mind-share matters.

>> The question in open source is always - who owns this? Whose is it? Who gets to say what happens next? Who are "we" and where is our community? I think it would be foolish to ignore those questions.

>An extended blog post may be helpful here, because this does feel like scaremongering and a few undeserved digs at Continuum. At least I've seen zero evidence of bad intentions on their part, quite the opposite actually.

Yes, it does appear to be true that raising these questions counts as scaremongering.   You may remember past email discussions where it became clear that it was considered rude and dangerous to raise the question of conflict of interest when discussing Travis' role in NumPy governance.  This is very unhealthy, and another reason why we can't afford not to discuss this question openly and in public.

rgommers:
> Note that I said "I was in practice obliged to install". Here we are thinking about what the future would look like if Anaconda became the only practical way to install the scipy stack. I don't consider that an unreasonable outcome if we push our users aggressively towards Anaconda. Mind-share matters.

We're very far from that now, and given the size of the Python community, the popularity and support of pip (it's dominant outside the scientific crowd), our continued efforts to support pip/wheels, and the determination of Linux and Macports/Homebrew packagers, it is very unlikely that we will reach that situation in the next 10 years.

GaelVaroquaux:
> We're very far from that now, and given the size of the Python community, the
> popularity and support of pip (it's dominant outside the scientific crowd), our
> continued efforts to support pip/wheels, and the determination of Linux and
> Macports/Homebrew packagers, it is very unlikely that we will reach that
> situation in the next 10 years.

More and more I get told that if I installed some packages via pip rather
than conda, it's no surprise that they are broken. This is not the case
for scipy. But it tells us that we are not that far from that situation
from a social perspective.

rgommers:
> More and more I get told that if I installed some packages via pip rather than conda, it's no surprise that they are broken. This is not the case for scipy. But it tells us that we are not that far from that situation from a social perspective.

It has never been a surprise that pip packages can break. ``pip`` (and ``easy_install`` before it) is a beta-quality tool. It has always been rough around the edges, and is continuously but slowly improving. 

The situation with the default distutils-sig tools is much better now than it has ever been (see for example [this blog post](https://glyph.twistedmatrix.com/2016/08/python-packaging.html)). So the implied opinion that Conda/Anaconda made things worse here is absolutely not the case.

I see this as: pip and predecessors were crap for many years, then a technically superior packaging tool comes along (also BSD-licensed and with completely open development by the way), and people start voting with their feet. They have a better experience with conda, and therefore tell you that if pip is broken, move to conda. There's nothing wrong with that. Competition is healthy.

rgommers:
> This is very unhealthy, and another reason why we can't afford not to discuss this question openly and in public.

Therefore I really encourage you to write that blog post, with a coherent argument that explains your views on community/political aspects. 

I tend to look at actual actions taken, how people and companies license their code/projects, etc. And at growing the community and attracting new users. And at technical quality and user-friendliness. Continuum and the packages they've open-sourced have been exemplary in all those aspects. So arguments like "binaries come from a company-controlled place and therefore are insecure, or bad for the community" make little real-world sense to me as small snippets throw into a discussion. I'd be interested to see the big picture view of someone who holds those opinions though.

takluyver:
I also find it surprising that you hear that, because most Python packages I know of are uploaded to PyPI by the authors, while conda packaging is a secondary step. What I typically see is developers making sure that the PyPI packages work, and leaving conda installation issues up to Continuum or conda-forge.

Especially with binary wheels now possible for all platforms (and the excellent tools Matthew, Nathaniel and others have made to build them), I'm definitely not worried about pip packaging for the foreseeable future.

And on the other side, the rise of conda-forge is starting to create a community-driven package ecosystem on the foundations of conda. It's not a done deal yet, but the trend is away from problematic corporate control of conda-land.

matthew-brett:
> I see this as: pip and predecessors were crap for many years, then a technically superior packaging tool comes along (also BSD-licensed and with completely open development by the way), and people start voting with their feet. They have a better experience with conda, and therefore tell you that if pip is broken, move to conda. There's nothing wrong with that. Competition is healthy.

First - I agree that, if Canopy and Anaconda had not done the work that they did to make Linux installation work in a general way, then we would not have the manylinux spec and Linux wheels.  So the work that the competition was doing, has been healthy.

Now, it seems to me, it is not so clear.    For the next few years, I anticipate conda and pip being in competition with each other, with the work on conda being a direct drain on the resources that could be going to improving pip, wheel coverage and pypi.   If conda wins in that competition, we (scientific Python) will be in a conda ghetto that may or may not have Continuum's installer and packages as key players.

matthew-brett:
> Therefore I really encourage you to write that blog post, with a coherent argument that explains your views on community/political aspects.

I will try to write that post.  In the mean-time, have a look at this thread for several examples where Continuum members are expressing surprise and disbelief at the idea that conflicts of interest could arise  : http://numpy-discussion.10968.n7.nabble.com/Governance-model-request-td41065.html .   Quoting from Nathaniel replying to Bryan Van de Ven:

"I understand where you're coming from, but I'd appreciate it if we 
could keep the discussion on a less visceral level? Nobody's personal 
integrity is being impugned, but it's the nature of this kind of 
governance discussion that we have to consider unlikely-and-unpleasant 
hypotheticals. It's like when you talk to a lawyer about a contract or 
a will or whatever: they'll make you think about all kinds of horrible 
possibilities, not because any of them are likely, but because sooner 
or later *something* will go wrong, and the point of having a 
contract/will/governance document is to provide some framework to 
handle whichever unlikely edge case does arise."

jakevdp:
@matthew-brett – the mailing list thread you quote is from a year ago; I believe concerns about longevity of Continuum's contribution may have been relevant then, but with the advent of conda-forge (and Continuum's effort over the last year to publish all their build recipes) I believe such concerns are not so relevant today.

jakevdp:
As for governance of conda-forge governance, that's something that's being hashed-out as we speak: we covered it at the last developer meeting, which is open to all: https://github.com/conda-forge/conda-forge.github.io/tree/14b820aad052f7b105a80283a3e12ecd67bbdb68/minutes

jakevdp:
Whoops, the most recent minutes haven't been merged yet: https://github.com/conda-forge/conda-forge.github.io/pull/237

matthew-brett:
@jakevdp - sorry - maybe I wasn't clear.  I wasn't thinking of the longevity of the Continuum contribution, but the potential for harm, or perception of that potential, with Continuum as a key player in the packaging ecosystem.  I was pointing out the worrying lack of concern about conflict of interest that has been a consistent feature of discussions with Travis and other Continuum members in the past.

jakevdp:
Ah, I see.

I have to say that given developments of the last year I mentioned above, I think there actually *is* little to be concerned about regarding conflicts of interest.

matthew-brett:
:)

rgommers:
> For the next few years, I anticipate conda and pip being in competition with each other, 

There's only partial overlap in capabilities, so that's only true for certain usages. There will remain a place for both.

> with the work on conda being a direct drain on the resources that could be going to improving pip, wheel coverage and pypi.

That is not how open source works. Even if those people that are now attracted to conda would be interested in working on pip/pypi, the state of development and decision making for pip/pypi is so bad that there's zero chance it could engage all those people productively.

Also: "competition is good until my favorite straggler has reaped the benefits, after that it's bad" is an odd view on the value of competition:)

GaelVaroquaux:
> Even if those people that are now attracted to conda would be
> interested in working on pip/pypi, the state of development and
> decision making for pip/pypi is so bad that there's zero chance it
> could engage all those people productively.

This is FUD, or at best an uninformed statement. The state of development
is quite good. The state of decision making is excellent. The pypa has an
excellent process of enhancement proposals with good discussions. A
discussion like the one surrounding
https://www.python.org/dev/peps/pep-0518/ is more thoughout than most I
have seen around packaging. It involves actors of very different
applications of Python (including our own Nathanael Smith), and with
different downstream distributions (Nick Coghlan works at Red Hat, and
does RPMs for Python). These days, the progress on Python packaging
infrastructure is excellent, albeit slow.

> Also: "competition is good until my favorite straggler has reaped the
> benefits, after that it's bad" is an odd view on the value of
> competition:)

I agree that competition is good. Competition via conda actually improved
pip :).

rgommers:
> This is FUD, or at best an uninformed statement.

Eh, I'm actively participating on distutils-sig and the pip repo. I'm probably one of the more informed people. And for every good example (some of the best indeed being driven by the scientific community) there's a bad one (or a couple).

GaelVaroquaux:
> Eh, I'm actively participating on distutils-sig and the pip repo. I'm probably
> one of the more informed people. And for every good example (some of the best
> indeed being driven by the scientific community) there's a bad one (or a
> couple).

Is this any different from conda? Disclaimer: I don't know where the
conda discussions happen.

rgommers:
Yes, it's very different. The energy that goes into conda - actions, hangouts, decisions - feels positive and to the point. distutils-sig is mostly frustrating. I've also seen several times that they blame new contributors for "not following through" when the core devs can't get their act together and decide which path to take to reach a goal.

rgommers:
To make a bit clear what the problem is: someone like Nathaniel can write a PEP-style document, shepherd the discussion on a not overly welcoming mailing list in the right direction, and write code that can be quickly merged. The contributors not able to that (i.e. the vast majority) are not helped and welcomed well in many cases. That can be due to the large workload the core devs are under, the large amount of people (including core Python devs) who offer drive-by opinions but no code, or just the culture. Either way, that is a real issue. The decision making process and governance for distutils-sig is very good, but the practice is not.

History and ratio of active to passive participants may be a big part of it. I've also heard people say numpy-discussion is not a nice/productive place, compared to for example the Cython mailing list. Could have similar reasons.

GaelVaroquaux:
I think that the reason of the differences is that pypa is trying to
catter for many more usecases and cultures. Reaching consensus is a very
hard difficulty. Hence the slower progress and less light-hearted
atmosphere.

On the other hand, in my opinion, Conda is a mono-culture by people who
all think the same, and are interested in only one pattern of usage. It
enables fast progress, to the cost of solving less problems (like: how to
integrate with another packager).

Maybe there is a lot of energy going in conda, but I find the attitude on
twitter and at conferences disrepectful of other people's point of view,
and often simply condescending on other packaging efforts. PyPA core devs
tend to do much better in this regard, trying to unite efforts. This is
what I really like about the energy going into PyPA.

rgommers:
> to the cost of solving less problems (like: how to integrate with another packager).

That was a conscious decision of scope right from the start. And it has been an effective one. There's also no need for conda and pip to be direct competitors with feature equality. Conda does things that are completely out of scope for pip (like packaging Python itself and non-Python packages), and vice versa (integrating with an existing Python install. In that way they're complementary, which is good.

> I find the attitude on twitter and at conferences disrepectful of other people's point of view, and often simply condescending on other packaging efforts.

I've never seen that, but that doesn't sound good.

matthew-brett:
On Wed, Sep 14, 2016 at 1:54 AM, Gael Varoquaux <notifications@github.com>
wrote:

> I think that the reason of the differences is that pypa is trying to
> catter for many more usecases and cultures. Reaching consensus is a very
> hard difficulty. Hence the slower progress and less light-hearted
> atmosphere.
>

Yes, I agree.   Of course you'll get less disagreement if you are trying to
solve the problem for your own subset of users.

I don't find pypa particularly dysfunctional - it's just there are a lot of
different interests that need to be resolved.  For example, pip and pypi
packaging has to negotiate with the buildout folks.  That's healthy, but of
course it's slow.

So, sure, you can disown all that stuff and do it on your own.  But the
result doesn't serve all the community, but a subset.

jakevdp:
> the result doesn't serve all the community, but a subset.

Yes! And from my experience teaching newcomers to the scipy universe, I would argue that they are a subset that conda serves *very* well. And, to bring the discussion full-circle, that's why I would support recommending conda on SciPy's installation page :smile:

matthew-brett:
Choosing speed and ease of decision-making over whole-community support is - in my view - a short-sighted choice and one that will cost us a great deal of wasted effort and frustration in the long term.

GaelVaroquaux:
> Yes! And from my experience teaching newcomers to the scipy universe, I would
> argue that they are a subset that conda serves very well.

That's your vision. My users often are set in a neuroimaging environment,
such as a lab-specific distribution, or the Python that comes inside a
neuroimaging software suite.

You are making an opiniated choice by recommending conda to people who
have no idea what it entails, and local support must then undo this
recommendation because it broke their environment.

jakevdp:
Well, I disagree. I suppose that fundamentally I'm still not quite understanding why you think supporting multiple tools for different use-cases is so damaging... perhaps I'll wait on your promised blog-post.

GaelVaroquaux:
> perhaps I'll wait on your promised blog-post.

I'm writing it. But I am trying to make it balanced, and sumarizing
different points of view, which takes time.

jakevdp:
> You are making an opiniated choice by recommending conda to people who
have no idea what it entails, and local support must then undo this
recommendation because it broke their environment.

Yes, and if we recommend something else it will also be an opinionated choice that will fail for certain users. And yet... making users wade through detailed pros and cons before doing a simple install will not result in clarity.

jakevdp:
> I'm writing it. 

Oh good! I was talking to @matthew-brett there... nice to know we'll be getting both your views on this :smile:

matthew-brett:
> Well, I disagree. I suppose that fundamentally I'm still not quite understanding why you think supporting multiple tools for different use-cases is so damaging... perhaps I'll wait on your promised blog-post.

Ralf and I agreed further up that pip and conda are in competition.   To me that competition seems obvious and direct, in the sense that we are recommending one over the other here.   In their practical use the overlap is much greater than the difference.  Clearly, the effort in supporting conda or conda-forge could be spent on improving pip packaging, and the other way round.

I don't think it needs a blog post to explain that, but I do think it needs a blog post to reflect on the long-term direction we're going.   I can't see a good outcome from here, with the current split in the community.

GaelVaroquaux:
> I don't think it needs a blog post to explain that,

Actually, from a cursory look, I had the impression that the PyData/scipy
commmunity was overly convinced of the benefit of a conda-only world.
Over the last month (since I gave my EuroPython keynote), I have talked
to many people that admit disagreeing with that view. The problem is that
people disagreeing self-censor themselves, as they don't like to appear
negative. Hence, it is useful to write such blog post, as long as they
are balanced and don't spread FUD but actual argument. The one of the
community split, that worries you as well as me, is a valid one.

matthew-brett:
I suppose the conda-only idea could be that, when conda is working as well as it can in the scientific community, the rest of the Python community will realize we were right all along and adopt conda as well.

In my experience, people like to think too much about the long-term direction of packaging, and only start to worry when they do think about it.

I guess that is a large part of what I am asking in this thread, is that rather than drifting down the easy path, we stop to think for a while about what these paths mean.

jakevdp:
> To me that competition seems obvious and direct

That depends what you're doing. I personally use conda because of features it has that pip **cannot do**, such as manage a complicated chain of non-python dependencies. This is why I tend not to view them as direct competitors, any more than I view ``pip`` and ``apt`` as direct competitors. And I'm not alone there: Nick Coghlin, for example, has largely the same views.

> the rest of the Python community will realize we were right all along and adopt conda as well.

That simply will not happen: pip and conda are different tools with widely different goals (albeit with a small region of overlap that's driving this argument). Have you read my recent blog post on the subject? You're repeating a lot of the myths I debunked there.

GaelVaroquaux:
>     To me that competition seems obvious and direct

> That depends what you're doing. I personally use conda because of features it
> has that pip cannot do,

Yet you advertise it instead of pip (here to install scipy, where is it
not needed). Hence the competition.

jakevdp:
Yes, I agree they compete within a small subset of tasks such as installing SciPy. But to extrapolate from there and imply that one could replace the other is not accurate.

matthew-brett:
@jakevdp - yes, I had read your blog post before this thread started.

matthew-brett:
The only uses I have personally seen of Anaconda and conda have more or less full overlap with pip.

Of course, if you want a packaging system that installs R as well as numpy, you're going to use apt or homebrew or conda, but, from my experience, that end of the conda spectrum is much less used than the simple scipy stack installs.

jakevdp:
> The only uses I have personally seen of Anaconda and conda have more or less full overlap with pip.

Great – just understand that is not the case for everyone.

jakirkham:
So I'm probably going to regret writing this, but I do want to add some information to the discussion that seems missing without simultaneously being pulled into a debate. As a result, I may not respond to a lot of comments/follow this discussion, but I do want to correct what appears to be a misunderstanding of `conda` and `conda`-users. In particular, I'd like to respond to this comment.

> On the other hand, in my opinion, Conda is a mono-culture by people who all think the same, and are interested in only one pattern of usage. It enables fast progress, to the cost of solving less problems (like: how to integrate with another packager).

I don't see this as accurate. In fact there is interest from `conda` users in integrating with other packaging systems. For instance...

* Building wheels - 
  * https://github.com/janschulz/conda-build-wheel
  * https://github.com/conda/conda-build/pull/1330
* Building RPMs
  * https://github.com/pelson/conda-rpms
* Using virtualenv/tox
  * https://github.com/conda-forge/staged-recipes/issues/1139
  * https://github.com/RonnyPfannschmidt-migration-tests/tox/issues/5

matthew-brett:
@jakirkham - perfectly reasonable comment.  I myself haven't seen much sign of unfriendly competition between pip and conda packagers, friendly and cooperative is much more common.   It's the advocacy that can get a bit heated.

ev-br:
TL;DR: I think we need all of: a one-sentence recommendation "Use Anaconda", and a collection of recipes for installing things with both pip/virtualenv and conda/conda-forge, and a list of pros and cons of various installation methods (what Ralf was writing), and long in-depth blog posts from Jake, Matthew and Gael.

My perspective is (quite similar to Jake's, I guess) is that I often teach computational-something courses to hard sciences students. Students typically have physics or maths background, they have prior experience with C or C++ (almost exclusively within Visual Studio on Windows machines only), and have zero experience with anything python related. Classrooms have collections of Windows desktops, and tech support is Windows only.

In these classes, the only realistic option on day one is "use Anaconda". And it's an Anaconda installer, not conda/miniconda/conda-forge. It's Anaconda or the class has to be taught with Matlab (or C++ on Visual Studio.)

A couple of weeks into the course, when they no longer struggle with basic syntax, some students need to install a package, or have a python installation on their laptops. At this stage, they are ready to learn that there is pip and conda, but they still need a recipe they can follow. Here we give them recipes for both conda/conda-forge and pip/wheels and they can make a random choice between the two. 

Once they get *that* going, they are ready to read about pros and cons of pip and conda. And then they can use links to long, in-depth, possibly opinionated blog posts from Jake, Matthew and Gael. (And of course, then there's a whole Internet full of opinions, but we only put links to high-quality ones in the scipy docs).

So my practical suggestion is that we should have an install instructions page with structured contents:

1. It starts with "If you're beginner, use Anaconda".
2. Then there's a collection of install recipes, for major operating systems, for both pip/virtualenv and conda/conda-forge
3. Then there's a section listing pros and cons
4. Then there are links to select blog posts.

This way, the page gets long, but this should not be a problem as long as different levels of assumed knowledge are clearly separated.

takluyver:
>  And it's an Anaconda installer, not conda/miniconda/conda-forge. It's Anaconda or the class has to be taught with Matlab (or C++ on Visual Studio.)

I was thinking about this point as I walked home. We often focus on the discussion between pip and conda as packaging systems, but the existence of Anaconda as an all-in-one installer is a big point in favour of that when we recommend something for new users.

If someone were to produce similar installers for the three major platforms based on the pip+wheels ecosystem, that would reduce a key reason that we recommend Anaconda as the easiest way to get started. Conversely, if someone were to make an installer from conda-forge packages (perhaps using [conda constructor](https://github.com/conda/constructor)), that would reduce the argument that corporate control is an issue for conda.

GaelVaroquaux:
> I was thinking about this point as I walked home. We often focus on the discussion between pip and conda as packaging systems, but the existence of Anaconda as an all-in-one installer is a big point in favour of that when we recommend something for new users.

IMHO, there are two things to be distinguished. The first one is whether
or not to recommend Anaconda for people who don't have a Python
installation. The second one is whether or not to recommend conda instead
of pip.

My usual point of view is that is people don't have a Python
installation, unless they are under a modern Linux distribution, I
recommend Anaconda [1].

But, I recommend to use pip unformely because:

1. Python tutorials will do so

2. Outside of the pydata community, the users will find more easily help
   with pip (eg by there local IT support).

3. There are more Python packages on pip than conda-forge

4. I can't think of a problem solved by conda that my users will face (I
   teach/distribute Python codes, not R, or something else).

So, my views would be 1) yes for Anaconda _provided_ that there is no
up-to-date Python install (it's quite infuriating to see that the
environments that we set up on desktops at work are void when non
technical users follow some instructions on Internet that tell that that
to get package foo they must install a new Python distribution), 2)
recommend pip, and if advanced people want to use conda, they can.

If I need to simplify things to the extreme, I say: Anaconda for Windows
and OSX, package manage for Linux, and pip for installing additional
packages.

[1] see for instance the nilearn installation instruction.
http://nilearn.github.io/introduction.html#installation

matthew-brett:
If I was doing a 2 day Python bootcamp for 200 people, with a mix of OSX, Windows and Linux, I would probably make an unhappy face and tell everyone to install Anaconda.

If someone wanted to have a quick go to see if they liked Python, I might recommend Anaconda.

When I'm teaching people who I'm expecting to work daily with their Python installation, I recommend homebrew / apt and pip, except on Windows : https://bic-berkeley.github.io/psych-214-fall-2016/preparation.html#installation .

It's really easy to build a one-shot installer based on pip : http://fullmonty.github.io . I haven't used this much because `pip install numpy scipy matplotlib jupyter` etc is not very hard, and I'm generally going to be encouraging people use the command line anyway.

rgommers:
> If I need to simplify things to the extreme, I say: Anaconda for Windows and OSX, package manage for Linux, and pip for installing additional packages.

That sounds about right. It's pretty close to the Jupyter install page, module the Linux package manager bit. The nilearn instructions are also very good and user-friendly (and I like the tabs). Much better than what we have on http://scipy.org/install.html now

takluyver:
> It's really easy to build a one-shot installer based on pip : http://fullmonty.github.io

Neat, I hadn't seen that before. :-)

Is it similarly easy to build installers for Windows and Linux? And is it easy to include Python itself in the installer? It's not the end of the world if people have to get two matching installers and run them both, but it's another opportunity for things to go wrong.

jakevdp:
One big consideration you left out, @matthew-brett: if you're using conda and the package/build you need is not available, it's trivial to fall-back to pip. If you're using pip and that happens, you have to build a new environment from scratch.

stefanv:
@jakevdp Do you know if it is safe to mix a pip install of numpy and a conda install of other packages? I saw a colleague install TensorFlow this morning, which brought in a numpy wheel.

stefanv:
Re the scikit-image installation instructions, we'll update those (the recommendation to avoid pip on Linux was before manylinux wheels were in place).

matthew-brett:
On Wed, Sep 14, 2016 at 1:54 PM, Gael Varoquaux <notifications@github.com>
wrote:

>
> If I need to simplify things to the extreme, I say: Anaconda for Windows
> and OSX, package manage for Linux, and pip for installing additional
> packages.


I would argue rather strongly for homebrew instead of Anaconda for OSX.

1) Any serious OSX user should have homebrew, for many more things than
Python;
2) Installing Python is then naturally `brew install python`;
3) Installing the scipy stack is `pip install numpy scipy jupyter pandas
matplotib sympy nose`;
4) There's no extra complication between them and the three programs they
already know; `brew`, `python` and `pip`.

matthew-brett:
On Wed, Sep 14, 2016 at 4:56 PM, Matthew Brett <matthew.brett@gmail.com> wrote:
>
> On Wed, Sep 14, 2016 at 1:54 PM, Gael Varoquaux <notifications@github.com>
> wrote:
>>
>>
>> If I need to simplify things to the extreme, I say: Anaconda for Windows
>> and OSX, package manage for Linux, and pip for installing additional
>> packages.
>
>
> I would argue rather strongly for homebrew instead of Anaconda for OSX.
>
> 1) Any serious OSX user should have homebrew, for many more things than
> Python;
> 2) Installing Python is then naturally `brew install python`;
> 3) Installing the scipy stack is `pip install numpy scipy jupyter pandas
> matplotib sympy nose`;
> 4) There's no extra complication between them and the three programs they
> already know; `brew`, `python` and `pip`.

To clarify, this advice is for someone new to Python, but serious in
intent.  For bootcamps or quick look-see, I would still recommend
Anaconda.

andyfaff:
I got burnt big time using Macports a few years ago. I seem to remember
lots of directories being created (couldnt keep track of them all) and I
had a lot of ports that didn't work. Dealing with PATH was annoying.
That experience put me off using things like homebrew.

Downloading Python frameworks was a pain, because it was really awkward to
fix paths so that the right python started up. When you want to rapidly
switch between python 2 and 3 that becomes  non starter. I also got caught
out because I you can't mess with system python.

In comparison I could download a miniconda script, all files go in a single
directory that I can nuke if need be. Easy switching between 2 and 3 and
easy to create environments. Plus you can use pip in a conda environment.

So I wouldn't say that brew is a 'must' for OSX. For other software that
brew offers, I normally just compile it myself, but perhaps I'm a bit weird
in that regard.

matthew-brett:
I'll rephrase then by saying that it's uncontroversial to recommend a serious OSX user install homebrew.

I haven't personally had any problems with the Python.org installers, I used to use those all the time, but because I now always install homebrew, the `brew, python, pip` combination seems the most natural.

jni:
I feel like I'm walking into a hornet's nest here, but I wanted to point out three use cases of conda that are absolutely not covered by pip, nor is there, apparently, any intention of ever supporting them. This is in quasi-direct response to @GaelVaroquaux's "I can't think of a problem solved by conda that my users will face (I teach/distribute Python codes, not R, or something else)."

### 1.

Installing non-system Python. The MPL instructions for OSX using pip are broken, unless `pip install` somehow now magically installs in user site-packages? The `get-pip` instructions issue a stern warning about messing with system Python, which somehow most pip proponents here seem to ignore:

screen shot 2016-09-20 at 9 54 32 am

This is *incredibly* important in Scientific Python, because *very quickly* in science, users want to install `master` branches of things, and packages without a PyPI distribution. It's very easy to end up with a broken system Python. I did it countless times before EPD and conda.

Additionally, system Python is almost always Python 2.7, for good reasons. There is no easier way to use Python 3.5 daily than with conda. Who wants to type `python3` when autocomplete goes to `python`? Ditto `pip3`. It's not surprising that many scientific Python *leaders*, let alone users, still use Py2.7 as their daily driver, given the discussions in this page. This speaks to @matthew-brett's "where we want to be as a community".

### 2.

Using different C compilers in different environments. Writing CPython often includes compiling Cython code. As you many of you know, gcc is aliased to clang in OSX. Again, this is not something I want to mess with, lest I break various dev tools that now assume this. But I recently had to test OpenMP support in Cython and I could just `conda install gcc` in a conda environment and it friggin just worked. 

### 3.

Installing non-Python dependencies. This is an obvious one but one that is often missed when most people in a discussion are Linux users. I'm bullish on JITs for the future of Python, and they tend to depend on complex lower-level packages such as LLVM, which might be trivial to install on Linux but are certainly not so on OSX and Windows.

Before I get pointed to Homebrew / MacPorts, I'll agree 100% with @andyfaff that these have done nasty things to my PATH in the past and so I have steered clear of them in recent years.

...

I'm impressed with PyPA, but the above use cases are important and, so long as they are deemed out of scope for pip, I'll remain a staunch conda advocate.

matthew-brett:
In order:

1 - absolutely you should not use System Python, which is why any sensible install instructions should suggest one of {homebrew, Python.org, miniconda, Anaconda}.   That's what we already have.  You might be complaining only about the lack of a standard alias of `python` to `python3`?   I'm really not sure that's a "where do we want to be as a community" issue.

2 - Despite doing an enormous amount of packaging work and compiling, I have very rarely needed gcc.  But, if you do need it, homebrew can install it, and much else.

3 - I don't myself ever use numba, but I wouldn't be surprised if this was easier in conda than for homebrew.  For the standard scipy-stack packages, all the dependencies ship with the wheels.

I am sure it's possible to run into problems with homebrew, as for any install method, but over the course of many mac installs in the last couple of years, including students installing for several courses, I have not had any problems with homebrew and the path.   My experience of macports has been less good, hence my recommendation for homebrew.

PR #189, opened 2016-11-15
MAINT: update links to scipy-user list for migration to python.org

rgommers:


rgommers:
This is a pretty trivial fix but important to get the links to the ML right. So I'll just go ahead and merge this.

PR #200, opened 2017-03-23
DOC: update mailing lists page for move to python.org infrastructure

rgommers:
**Merge only after the move has actually taken place!**


rgommers:
Lists moved, so merging.

ev-br:
Yay! Thanks to all involved

PR #201, opened 2017-03-24
MAINT: fix "make upload"

rgommers:


rgommers:
- new.scipy.org not relevant anymore
- ``make upload`` now runs without any errors. that never happened before. looks like we don't need the ``fixperm-scipy_org`` script anymore, permissions look fine as is.

pv:
Why the --chown option would work is unclear if you are not uploading as 
root.
.
It's unclear if the upload will work for other users than you, without 
fixing the permissions to group-writable.

pv:
But let's try to push the move to S3 or github pages ASAP, so that all 
this goes away.

rgommers:
> Why the --chown option would work is unclear if you are not uploading as root.

Not 100% sure but it does. Didrik changed something for ```www-data`` with ``setuid``/``setgid`` to make this work.

> But let's try to push the move to S3 or github pages ASAP, so that all this goes away.

That'd be good indeed. Enthought people are looking into it.

This is still a useful PR to merge I think, fixes several issues aside from the ``chown``.

PR #202, opened 2017-03-27
Add SymPy paper to citing.rst

Sumith1896:
We recently published and the paper can be found here https://peerj.com/articles/cs-103/
We would love it to be included here.


rgommers:
LGTM, merging. Nice paper!

Sumith1896:
Thanks @rgommers :)

PR #203, opened 2017-03-28
Update mailing lists page.  Remove links to Gmane, add Stackoverflow.

rgommers:
Note that there is no scipy-dev Nabble forum, so we lost easy
searchability for scipy-dev.  We may look into creating such a forum
soon.

### DIFF
--- a/www/scipylib/mailing-lists.rst
+++ b/www/scipylib/mailing-lists.rst
@@ -8,16 +8,17 @@ receive user support.
 
 **Searching a list** is a great way to get your questions answered without
 actually signing up for a list. Follow the **Read/Search** link for a nice
-interface from Gmane_. Or, browse the old fashioned
-way at the official **Archives** link.
+interface from Nabble_. Or, browse the old fashioned way at the official
+**Archives** link.
 
 To post to a list, you have to be on it. If you want to see **how much
-traffic** a list gets, there is a chart on the Gmane page, or just go to the
-list archive and take a look. You can opt to receive messages in digest form.
-Groups of about 10 messages are sent in one email, with a table of contents at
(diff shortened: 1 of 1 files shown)


pv:
I would hold it until it becomes clear that Nabble admins update the mailing list addresses. I'm not sure if this will happen: the numpy/scipy forums were created by someone (username "Timmie" there), and I'm not sure if the site admins will touch requests to change someone's forum settings. 

The mailing lists probably will become searchable via Google et al. anyway, although it may take some time before they get crawled --- the scipy.org archives can be searched:
https://www.google.fi/search?q=site:http://mail.scipy.org/pipermail/scipy-dev/+running+tests

rgommers:
They're both updated, here are messages from today:

http://numpy-discussion.10968.n7.nabble.com/Fwd-SciPy2017-Sprints-FinAid-for-sprint-leaders-core-devs-td44132.html

http://scipy-user.10969.n7.nabble.com/SciPy-User-Announcement-scikit-image-0-13-0-tt20871.html

pv:
I'm not sure if that's the case. Nabble is probably subscribed to the 
list as usual, and the subscriber list was transferred by us.

pv:
Anyway, if you think it works, please go ahead.

rgommers:
Either way it's much better than a dead link. I'll merge and set a reminder for myself to check in a week or two that the archive is complete. If not, we may just want to create new forums on Nabble, or search for an alternative.

rgommers:
Just checked: both Nabble forums are still updating

http://scipy-user.10969.n7.nabble.com/
http://numpy-discussion.10968.n7.nabble.com/

PR #205, opened 2017-04-04
MAINT: update location to push built site to:

rgommers:
served from ``/var/www/html`` now


rgommers:
Just made this work with @dpinte, so going to merge straight away.

PR #209, opened 2017-05-11
Add CMA as python optimiser in topical-software.rst

ElieGouzien:
Just a good specialized optimization software in python.


rgommers:
Thanks @ElieGouzien, merged. Sorry it took a while.

PR #211, opened 2017-06-05
Fix URLs for Veusz

jeremysanders:
Please update the homepage for Veusz. Also mentions creates PDF, SVG and bitmap output.


rgommers:
Merged, thanks @jeremysanders

PR #224, opened 2017-09-23
topical-software: update links to PyPi, drop central

pv:
Point pypi search links to https://pypi.org instead of
https://pypi.python.org/pypi -- the latter page nowadays says it's
deprecated interface, and it doesn't seem any longer possible to search
inside a specific category.

Drop links to scipy central.

### DIFF
--- a/www/topical-software.rst
+++ b/www/topical-software.rst
@@ -7,15 +7,14 @@ categorized by scientific discipline or computational topic.  It is intended to
 be exhaustive.  If you know of an unlisted resource, see About This Page,
 below.
 
+Additional useful software packages can be
+found on the `Python Package Index <https://pypi.org/>`__
+especially in its `Science/Research category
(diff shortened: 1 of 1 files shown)


rgommers:
I don't really mind this change (Warehouse / pypi.org are the future), but https://pypi.python.org/pypi?:action=browse&c=40 works fine right now, and the new page has a big red banner saying

    This is a pre-production deployment of Warehouse. Changes made here affect the
    production instance of PyPI (pypi.python.org).

Where do you see the deprecation?

pv:
The link https://pypi.python.org/pypi/ shows a message about deprecation --- apparently, if you drop the final `/` from the URL it gives a different page (which is what confused me here).

The issue for me on the old pypi is that the search box doesn't respect the selected categories.

Anyway, if the red warning on pypi.org is too distracting, maybe we want to wait for the moment.

rgommers:
Ah yes, see it now. The new search interface is much nicer. Let's go for making the change now I'd say.

rgommers:
In it goes, thanks Pauli.

PR #231, opened 2017-09-28
Remove Pyaire

bmcage:
Pyaire is prealpha package with no development since 2005


rgommers:
Thanks again, merged.

PR #234, opened 2017-10-07
Change installation instructions to promote pip + remove references to "scipy stack"

pv:
Update the installation instructions:

- Remove references to the "scipy stack" (following #218)
- Promote using pip also on Windows

We may want to wait until scipy 1.0 is out in order to have the windows wheels...

### DIFF
--- a/www/install.rst
+++ b/www/install.rst
@@ -1,139 +1,108 @@
-==========================
-Installing the SciPy Stack
-==========================
+===================
+Installing packages
+===================
 
-These are instructions for installing :ref:`the full SciPy stack
-<stackspec>`.  For installing individual packages, such as NumPy and
-SciPy, see :ref:`individual-packages` below.
+These are general instructions for installing packages in the SciPy
+ecosystem.
 
 Scientific Python distributions
 -------------------------------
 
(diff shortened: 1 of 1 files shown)


xoviat:
Also, there is a pip packaged called `scipy-stack`. So after 1.0 is released, you can download Python from Python.org and then type `pip install scipy-stack` and everything will work (Jupyter, IPython, numpy, scipy, etc...).

pv:
ready?

xoviat:
Yes but it has no code, only dependencies. You could ask for maintainer access.

xoviat:
@David-OConnor

pv:
Well, given that we decided the whole concept is not necessary any more, I think we won't re-add it in this PR.

xoviat:
BTW, how do you auto cancel the old PR builds? I could not find that option. 

But yes, looks okay.

David-OConnor:
Re the scipy-stack convenience package: There's not much to maintain! It should install properly on Windows, now that Scipy 1.0.0 is out.

pv:
Ok, I'll just merge it. If there's something to change we can mop it up later.

PR #235, opened 2017-10-11
DOC: building: windows

xoviat:
Re-filed against the correct repository.

@dalito or @dhermes Please take over this PR and fix the RST. Also, I think "Cygwin" should be MSYS2.


dhermes:
@xoviat My only laptop with a working Windows install also happens to be my "productivity" laptop (currently booted to Ubuntu). I'll **try** to take a stab at building this in the evening sometime this week.

dalito:
Thanks. I'll check/test at the next weekend.

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
 .. _OpenBLAS: https://github.com/xianyi/OpenBLAS
 .. _LAPACK: http://www.netlib.org/lapack/
 
+
+Building the Released SciPy
+---------------------------
+
+This section provides the step-by-step process to build the released scipy. If you want
+to build completely from source, you should estimate at least three hours to build all
+libraries and compile SciPy. Feel free to stop and inspect any step at any time, but
```
matthew-brett: Maybe rephrase this - if people do run into trouble this sentence will make them feel particularly bad.

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+we're mostly just going to breeze through this. More in-depth documentation is provided 
+below.
+
+First, download all of the following software concurrently:
+
+1) MinGW-w64 from https://mingw-w64.org
```
matthew-brett: This will lead them to the main site, but not to a particular download, I believe.  Specify a particular distribution for them to try / mention the different builds?

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+2) Microsoft Visual Studio 2015 or 2017 Community Edition (available from Microsoft)
```
matthew-brett: Full or command line tools? (I think full is 11GB minimum).

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+3) git from https://git-scm.org/
+4) MSYS2 from http://www.msys2.org/
+5) Python **with pip** from https://python.org/
```
matthew-brett: They might get confused about "with pip" here - it gets installed by default.  Maybe "; if you customize the install, make sure you leave the option to install "pip" checked".

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+
+While that's downloading let's outline our plan of action. We're going to:
+
+1) Build OpenBLAS using the MinGW-w64 compiler suite, using the build system from MSYS2
```
matthew-brett: I guess another option is to point them to the pre-built blas and build instructions to build it if they want.

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+2) Copy the associated libraries to the `Python\\Lib` directory
+3) Build SciPy with a hybrid MinGW-w64 toolchain
```
matthew-brett: Maybe expand to "build scipy using MSVC++ for the C code and mingw-w64 for Fortran".

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+
+Once git installed, open a terminal in your favorite directory and type the following:
```
matthew-brett: Grammo - "One git is installed, "

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+
+.. code:: shell
```
matthew-brett: I think you need a blank line after the `.. code` line.

www/scipylib/building/windows.rst
matthew-brett: I think you mean a git bash terminal here?   I can't remember whether msys git puts git on the Windows PATH by default, but it's certainly an option not to have it on the path.  So `cmd` might not work here.

Can't you install git via msys2?  Wouldn't that be simpler overall (don't need two installers, two terminals)?

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+   git clone https://github.com/matthew-brett/build-openblas.git
+   cd build-openblas
+   git submodule update --init --recursive
+
+Now let's build OpenBLAS. Open a MSYS2 terminal and change directories to the
+`build-openblas` folder. To make sure that we're ready to build, type the following in
+the terminal:
+
+.. code:: shell
```
matthew-brett: Blank line after - I believe.

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+   git
```
matthew-brett: I guess this is git from the git installer, present here because it is on the Windows PATH?

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+points to the version that you want to build. Make sure that the architecture is
+correctly set. And after you've made sure of that, build OpenBLAS.
+
+.. code:: shell
+    .\build_openblas.sh
+
+Building OpenBLAS is extremely problematic and may fail if your system is not correctly
+configured. Your build may fail after a few hours and you may have to restart it after
+fixing an undocumented problem. OpenBLAS builds can also fail silently and produce
+and incorrect binary. Please, if you have any issues, report them to
```
matthew-brett: an incorrect binary.

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+https://github.com/scipy/scipy.org so that we can save the next person's time.
```
matthew-brett: Point to the issues specifically - https://github.com/scipy/scipy.org/issues

Is that where Windows building reports should go?

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+
+After you've build OpenBLAS, there will be an `openblas.a` file somewhere on your system.
```
matthew-brett: built

matthew-brett: Don't we know where it will be on their system?  We should at least point them in the general direction.

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+If you don't have that file, you'll probably need to find out what happened and then
+build OpenBLAS again. If you have that file, then you may have built OpenBLAS correctly.
+Proceeding on that assumption, let's build SciPy. Copy `openblas.a` to the
+`Python\\Lib` directory and open a (non Cygwin) terminal. Then run the following commands.
```
matthew-brett: Do they know what a Cygwin terminal is?  Does msys2 have its own Python?  Can they depend on an msys2 finding the correct pip / python?  Or do they have to use a cmd / powershell terminal?

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+Please note that this is a simpler procedure than what is used for the official binaries.
+**Your binaries will only work with the latest numpy version**. For building against
+older Numpy versions, see Building Against an Older Numpy Version. After you have Numpy
+installed, run the following command in the (non Cygwin) termial.
+
+.. code:: shell
+    gfortran
+
+If gfortran is not on your path, then you will need to add it. Note that this may be
+different from the MSYS2 terminal. Assuming that you have set up everything correctly
```
matthew-brett: What does this mean - "may be different from the msys2 terminal." ?

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,109 @@ information than this document will provide.
...
+run the following.
+
+.. code:: shell
+    git clone https://github.com/scipy/scipy.git
+    cd scipy
+    pip wheel -v -v - v .
+    python runtests.py --mode full
+
+Congratulatations, you've build SciPy!
```
matthew-brett: built

pv:
MSYS2 ships with both mingw-w64 compilers and git, so it's probably the 
best option for installing both.

pv:
msys2 Python is iiuc built with mingw-w64, so it's not the right one for 
building scipy.

xoviat:
> msys2 Python is iiuc built with mingw-w64, so it's not the right one for building scipy.

That's correct. If I understand correctly, we need the MSYS2 build system (to shell out to gfortran / gcc) to build a performant OpenBLAS.

@matthew-brett I understand your concerns about the document (it's not really ready to be merged). However, I'm going to wait until we have some people who have actually done the procedure and then we can address the feedback all at once.

ilayn:
I've tried to follow your example but after installing msys2 things went sideways and I couldn't complete the instructions. The main issues are: 

  - In cygwin, it's not clear at all what to install or which package set. 
  - in msys, I'm pretty sure most people will struggle to `cd` out of msys directory. It took me a while to understand that I need to use `cd /c/` to get to the drive. 
  - after installing mingw64, I've found that msys package manager also has mingw installation options. I've followed https://github.com/orlp/dev-on-windows/wiki/Installing-GCC--&-MSYS2 page and it went pretty OK and all git, make, gcc, gfortran was ready on a msys-Mingw64 terminal which I was able to run build_openblas.sh

I'll add more if I can finish the build cycle. I still have some problems with the build.

xoviat:
@ilayn First, there should be no Cygwin. I know that it says that in this PR but that's a mistake. It should be MSYS2/MinGW. 

Second, what I've done is I've shifted the focus to building OpenBLAS first without worrying about building SciPy until OpenBLAS has been built. Do you have the commands that will install the packages required to build on MSYS2?

ilayn:
OK 8 hours and I'm at where @matthew-brett also got blocked https://github.com/scipy/scipy/issues/7597#issuecomment-315078397. Hopefully tomorrow, I'll get it working by reading the rest of that thread.

xoviat:
What were the commands to install the MSYS2 packages?

ilayn:
> Second, what I've done is I've shifted the focus to building OpenBLAS first without worrying about building SciPy until OpenBLAS has been built. Do you have the commands that will install the packages required to build on MSYS2?

Oops missed your comment, the tab didn't refresh. Yes in a nutshell I know roughly what I did (checking the history on the terminal). I'll try to summarize.

ilayn:
@xoviat  After performing the regular instructions until `pacman -Syuu` gives nothing to update, I've run the following and installed these

    pacman -S --needed base-devel 
        mingw-w64-i686-toolchain
        mingw-w64-x86_64-toolchain
        git subversion mercurial
        mingw-w64-i686-cmake mingw-w64-x86_64-cmake

xoviat:
Thanks. Also, you should know that you probably won't be able to compile SciPy be reading that issue thread. Reading it again, I don't think I communicated very well (but in fairness, I was deep in debugging build issues).

xoviat:
I will try to update the PR to give better build instructions.

ilayn:
No worries I have to get acquainted with this workflow anyways. I'll see what I can do.

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,172 @@ information than this document will provide.
...
+After you've built OpenBLAS, there should be an `openblas.a` file somewhere on your system.
+If `OPENBLAS_ROOT` was set to `C:\\opt`, then you might see a line like this in the MSYS2
+terminal:
+
+.. code:: shell
+   Copying the static library to /c/opt/64/lib
+
+If you see that line, then you might have built OpenBLAS correctly, even if other failures
+occurred. Look in that folder for `openblas.a`. If the file isn't there, then poke around and
+try to find the file elsewhere. If you don't have that file, you'll probably need to find out
```
xoviat: @ilayn Build procedure has been reworked on focus on building OpenBLAS first and emphasizing the fact that the failures related to `lib.exe` that you experienced don't matter.

ilayn: @xoviat  No problem. I'm just adding the extra issues that comes from the `build_openblas.sh`. Note that the file name turns out as `libopenblas_5f998ef_gcc7_2_0.a` after this workflow (with postfix). So that also needs to be handled. Also I think you need extra empty lines after `..code: shell::` directives for them to be rendered.

ilayn:
Now I've read it again and I think the packages are a bit too many. We don't use mercurial or subversion anyways. So probably it is enough as below

     pacman -S --needed base-devel mingw-w64-i686-toolchain mingw-w64-x86_64-toolchain mingw-w64-i686-cmake mingw-w64-x86_64-cmake git

ilayn:
@xoviat In your comment https://github.com/scipy/scipy/issues/7597#issuecomment-315122714, you point to a code fragment which is now nonexistent. Could you please point me to it if possible? I think I roughly understand the problem but I don't which the `.def` files I should use.

pv:
You should not need any .def files. That part of the discussion in gh-7597 is about a failed, different approach.

ilayn:
Yes, I'm just trying to replicate the efforts to see whether I can keep up with the discussion. I'm not able to go through the DLL branch either so just following the steps :)

xoviat:
Did you follow the instructions in this pull request? Do you have the msvc build tools installed? Do you have mingw-w64 installed (not MSYS2)? Do you have python from python.org installed?

dalito:
@xoviat, which pre-built toolchain from http://mingw-w64.org/doku.php/download is suggested? I selected "Mingw-builds".

I have not yet completed all steps (the openblas build does really take time...). I will give feedback at the end. So far I had only minor problems to follow the guide.

xoviat:
I used the "mingw builds" option. Should probably clarify that.

xoviat:
Any update? Is the OpenBLAS build still going?

www/scipylib/building/windows.rst
```diff
@@ -32,6 +32,203 @@ information than this document will provide.
...
+You don't necessarily need to build in that particular location, but it should be somewhere
+reasonable. Now clone the repository required to build OpenBLAS:
+
+.. code:: shell
+
+   git clone https://github.com/matthew-brett/build-openblas.git
+   cd build-openblas
+   git submodule update --init --recursive
+
+Now let's build OpenBLAS. To make sure that we're ready to build, type the following in
```
matthew-brett: Repeating my comment before - there will be some people here (maybe most) who don't want to build OpenBLAS - how about pointing them to the built binary / appveyor repo, and putting these instructions in another page?

pv:
I would recommend here:

- We don't have bandwidth to maintain explanation of multiple 
configurations, so best to limit this to "one true way" that works.

- Scrap all of the old text on the page. (If there's a link to 
instructions maintained by Intel on how to do MKL builds, we can keep 
that link.)

- Keep the text on the page to minimum (skip explanations of motivations 
and alternative ways to do things). There's too much text IMHO 
currently, in comparison to how complicated the setup really is.

- Separate the instructions to parts, 1) install all compilers etc, 2) 
build openblas, 3) build scipy. People who already have what they need 
can skip reading.

- I'd prefer we explain using `site.cfg` for openblas rather than 
copying the lib files.

- Cygwin should not be mentioned.

- Keep the number of different websites to download stuff from to 
minimum --- I'd recommend to just install the toolchain + git from MSYS2.

- Why does the text suggest to not use the msys2-provided mingw 
toolchain --- the msys2 toolchain works fine for me?

www/scipylib/building/windows.rst
xoviat: Okay, but the ATLAS builds won't work (I think you may have suggested them).

matthew-brett: No, I meant the binary builds of OpenBLAS.

dalito:
Sorry, I only have time in the evenings. The openblas-build was successful (on German Win10 Pro, MSYS2 freshly installed).

For installing msys2 we should point to the official docs in the MSYS2 wiki: https://github.com/msys2/msys2/wiki/MSYS2-installation
BTW, I was surprised that I could not find a https-download link for msys2.

The pacman command needs line continuation signs "\\"
```
pacman -S --needed base-devel \
    mingw-w64-i686-toolchain \
    mingw-w64-x86_64-toolchain \
    git subversion mercurial \
    mingw-w64-i686-cmake mingw-w64-x86_64-cmake
```
Unclear: "Open a MSYS2 terminal and type the following:"
In "MSYS2 MSYS" terminal gfortran and gcc are not available. One needs to open "MSYS2 MinGW 64-bit" (or "MSYS2 MinGW 32-bit" for a 32-bit build)

What is meant with: "Make sure that the architecture is correctly set." - Just 32/64bit or something else like processor type?

The build command failed:
```
$ build_openblas.sh
bash: build_openblas.sh: Kommando nicht gefunden.
```
Do I have to make the sh script executable in MSYS2? Anyway I could start the build with :
```
$ sh build_openblas.sh
cygpath: can't convert empty path
HEAD ist jetzt bei 5f998efd... Merge pull request #1247 from martin-frbg/revert-cgroups-pr
```
The missing "cygpath" env. var did not stop the build process.

xoviat:
@dalito Do you have the `openblas.a` file? It doesn't matter then that the script failed.

ilayn:
@dalito  you just type `./build_openblas.sh` I think the slash is wrong in the document.

xoviat:
@ilayn Did you get further in the build?

ilayn:
@xoviat No unfortunately I'm still trying to get past the missing smybol issue. I've also tried to mimic the DLL copying following the appveyor yml file but that also didn't past the same quadpack missing symbol issue. 

I have 2017 Community Edition on Windows 10, Python 3.6.2, I have the MSYS2 and also the standalone installation of MINGW-W64 `x86_64-7.1.0-posix-seh-rt_v5-rev2`. All are on the path.

xoviat:
Is the `openblas.a` file in the `lib` directory?

ilayn:
It is indeed yes.

xoviat:
Is numpy upgraded to the latest version?

ilayn:
Yes I've tried both 1.13.1 and 1.13.3. It always gets choked here. 

```
   Creating library build\temp.win-amd64-3.6\Release\scipy\integrate\_quadpack.cp36-win_amd64.lib and object build\temp.win-amd64-3.6\Release\scipy\integrate\_quadpack.cp36-win_amd64.exp
mach.lib(d1mach.o) : error LNK2001: unresolved external symbol _gfortran_st_write
mach.lib(d1mach.o) : error LNK2001: unresolved external symbol _gfortran_st_write_done
mach.lib(d1mach.o) : error LNK2001: unresolved external symbol _gfortran_stop_numeric
mach.lib(d1mach.o) : error LNK2001: unresolved external symbol _gfortran_transfer_character_write
mach.lib(d1mach.o) : error LNK2001: unresolved external symbol _gfortran_transfer_integer_write
mach.lib(d1mach.o) : error LNK2001: unresolved external symbol _gfortran_stop_string
build\lib.win-amd64-3.6\scipy\integrate\_quadpack.cp36-win_amd64.pyd : fatal error LNK1120: 6 unresolved externals
error: Command "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.11.25503\bin\HostX64\x64\link.exe /nologo /INCREMENTAL:NO /LTCG /nodefaultlib:libucrt.lib ucrt.lib /DLL /MANIFEST:EMBED,ID=2 /MANIFESTUAC:NO /LIBPATH:C:\opt\64\lib /LIBPATH:"C:\Program Files\mingw-w64\x86_64-7.1.0-posix-seh-rt_v5-rev2\mingw64\lib\gcc\x86_64-w64-mingw32\7.1.0" /LIBPATH:"C:\Users\Ilhan Polat\AppData\Local\Programs\Python\Python36\libs" /LIBPATH:"C:\Users\Ilhan Polat\AppData\Local\Programs\Python\Python36\PCbuild\amd64" /LIBPATH:build\temp.win-amd64-3.6 /LIBPATH:"C:\Users\Ilhan Polat\AppData\Local\Programs\Python\Python36\libs" /LIBPATH:"C:\Users\Ilhan Polat\AppData\Local\Programs\Python\Python36\PCbuild\amd64" /LIBPATH:build\temp.win-amd64-3.6 /LIBPATH:"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.11.25503\Lib\x64" /LIBPATH:"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.11.25503\ATLMFC\Lib\x64" /LIBPATH:"C:\Program Files (x86)\Windows Kits\10\lib\10.0.16299.0\um\x64" /LIBPATH:C:\WINDOWS\Microsoft.NET\Framework64\v4.0.30319 /LIBPATH:"C:\Program Files (x86)\Windows Kits\10\lib\10.0.16299.0\ucrt\x64" quadpack.lib mach.lib libopenblas.lib libopenblas.lib /EXPORT:PyInit__quadpack build\temp.win-amd64-3.6\Release\scipy\integrate\_quadpackmodule.obj /OUT:build\lib.win-amd64-3.6\scipy\integrate\_quadpack.cp36-win_amd64.pyd /IMPLIB:build\temp.win-amd64-3.6\Release\scipy\integrate\_quadpack.cp36-win_amd64.lib" failed with exit status 1120
```

xoviat:
If that was the case, I'm so sorry for wasting your time. I thought that because the latest numpy had come out before the changes were merged into master that they would be in the version on pypi. Then I realized that numpy doesn't work like that.

ilayn:
No problem at all. There are too many things to consider and it is hard to anticipate all the details. 

However, I get the same result with NumPy 1.14.0.dev0+77f9540

ilayn:
I don't know if related but numpy install has also finished with 

```
Installing collected packages: numpy
  Running setup.py install for numpy: started
    Running setup.py install for numpy: still running...
    Running setup.py install for numpy: finished with status 'done'
Exception:
Traceback (most recent call last):
  File "c:\users\ilhan polat\appdata\local\programs\python\python36\lib\site-packages\pip\basecommand.py", line 215, in main
    status = self.run(options, args)
  File "c:\users\ilhan polat\appdata\local\programs\python\python36\lib\site-packages\pip\commands\install.py", line 342, in run
    prefix=options.prefix_path,
  File "c:\users\ilhan polat\appdata\local\programs\python\python36\lib\site-packages\pip\req\req_set.py", line 784, in install
    **kwargs
  File "c:\users\ilhan polat\appdata\local\programs\python\python36\lib\site-packages\pip\req\req_install.py", line 922, in install
    with open(inst_files_path, 'w') as f:
OSError: [Errno 22] Invalid argument: '"c:\\users\\ilhan polat\\appdata\\local\\programs\\python\\python36\\Lib\\site-packages\\numpy-1.14.0.dev0+77f9540-py3.6.egg-info\\installed-files.txt'

```

xoviat:
Check explicitly whether you have this line on your system:
https://github.com/numpy/numpy/blob/master/numpy/distutils/fcompiler/gnu.py#L438

ilayn:
That's also a check.

xoviat:
I'm stumped. The root of the problem is that Numpy is telling MSVC to link a static library full of gfortran object code, which is impossible. This is not something that Numpy should do and it's a complete dead end (eg. don't go hunting down symbols because it's a waste of time).

How about posting your entire build log as a gist?

ilayn:
Here it is https://gist.github.com/ilayn/d804b8cae0736cb580ccdb914c4abf99

ilayn:
This time the error is different if I run it inside a MINGW64 console from MSYS2. I'll add the standard console output in a bit.

xoviat:
Uh..oh. Numpy is picking up the "openblas.lib" and "libopenblas.lib". Make sure you remove all of those files from your system.

ilayn:
OK apparently something has changed because now it is not finding the quadpack libraries in the C:\opt64\lib directory which is clearly the wrong place to look for 

https://gist.github.com/ilayn/3735c7b44c51aa34c3540875f634a9ec

xoviat:
This line is a good thing:

> LINK : fatal error LNK1181: cannot open input file 'openblas.lib'

However, better would be for it to find the `openblas.a` file that you made.

xoviat:
You can see where it's looking for the files;

>   libraries lapack not found in ['C:\\Users\\Ilhan Polat\\AppData\\Local\\Programs\\Python\\Python36\\lib', 'C:\\', 'C:\\Users\\Ilhan Polat\\AppData\\Local\\Programs\\Python\\Python36\\libs']

xoviat:
More bad:

>   FOUND:
    libraries = ['libopenblas', 'libopenblas']
    library_dirs = ['C:\\opt\\64\\lib']
    language = c
    define_macros = [('HAVE_CBLAS', None)]

  FOUND:
    libraries = ['libopenblas', 'libopenblas']
    library_dirs = ['C:\\opt\\64\\lib']
    language = c
    define_macros = [('HAVE_CBLAS', None)]

ilayn:
I think I found the problem. This is all from a site.cfg nonsense I've tried a while back. And it is picking up the wrong openblas build because I've copied from the template from the openblas build. Things went back to normal after a git clean -xdf since it removed that file too. Now the build finished with `copying ..... .dll` statements. I guess that's a success ?

xoviat:
Yes.

xoviat:
How can we make clear that the lib files should not be used?

ilayn:
Hmm I've installed it and now it's giving me the error 
```

import scipy as sp
Traceback (most recent call last):

  File "<ipython-input-2-7e642d76ec1e>", line 1, in <module>
    import scipy as sp

  File "c:\users\ilhan polat\appdata\local\programs\python\python36\lib\site-packages\scipy\__init__.py", line 61, in <module>
    from numpy._distributor_init import NUMPY_MKL  # requires numpy+mkl

ImportError: cannot import name 'NUMPY_MKL'
```

But one step at a time. Apart from that, the build was a success ! WOOT 😄 

I think we have to make sure that after creating the archive file `.a` we instruct people to delete the folder of the OpenBLAS build

xoviat:
@ilayn Just delete all of the contents of `distributor_init.py`.

xoviat:
Updated with a stern warning about deleting the lib files.

ilayn:
OK cleaned up everything rebuilt numpy and scipy run a few tests from a few modules and apparently all works. Thank you for the patience :)

xoviat:
Others: more feedback? More reviews?

Also, responding to @pv:

ilayn:
As far as I understand from what I've done and apart from the technical perspective, I think the best way is the most documentable and reproducable way. 

For example, if I clean the repository site.cfg is also gone. So if I make a mistake in the build then I'll have to recreate it for every cleanup. 

I think if separate things into sections and change the narrative to *parts of a machinery* rather than sequential steps to reach to the end, here will be no problem (I think). Things to download are essentially 

  - MSYS2
  - MINGW-W64
  - SciPy Repository clone

Essential things to do : 

  - add compilers to path
  - update msys2 system and install its packages
  - modify the `export` vars in msys2 system
  - find the archive file rename it to `openblas.a` and nothing else, copy it to `\Lib` and delete all traces of the openblas build e.g. deleting `C:\opt` after taking the archive file out
  - Probably add a `python setup.py config` just to see the situation

and that should roughly be it. If modifications to site.cfg are involved then things will be more complicated in my opinion. If we use the site.cfg path then probably `build_openblas.sh` should create it on-the-fly.

pv:
@ilayn: MINGW-W64 is not necessary to download, the MSYS2 compilers are 
fine.

Not clear to me how removing MSYS2 and needing to install mingw-w64 
compilers again avoids confusion. IIRC, scipy builds fine with MSVC from 
the MSYS2-64 command prompt, without any additional configuration. 
Switching between powershell and msys2 (which you use for building 
openblas) is probably not necessary.

pv:
The only thing to be careful about is to use the Python.org interpreter, 
i.e.,
```
"/c/Program Files/Python36/python" setup.py build
```
I think msys2 does not install Python by default, so it's not so easy to 
get confused.

ilayn:
msys2 comes with Python2 but I needed Python3 which I already had on the system.  At the end, we have py2, py3 in msys2 and also py2 and/or py3 installed in windows. As you mention, that was the confusion I had in mind. But in general I think both have a tradeoff in terms of complexity and confusion. 

What I have found nice is that if you also include the mingw-w64 compilers, you have systemwide usable compilers on the native console instead of going into msys2 again as a bonus after this adventure. 

For Win10, I think this path can be made independent from MSYS2 via Windows Bash but I don't know if it is worth the hassle since win7, win8 users won't benefit from it.

pv:
Having system-wide compilers on PATH is bad when you want to compile for 
both 64-bit and 32-bit Python, best avoided.

Are you saying the msys2 installer pre-installs a version of Python? 
IIRC this is not the case. You do not want or need to install any 
version of Python inside msys2, since you are building for the 
Python.org python, not for msys2 python (which afaik is compiled with 
mingw).

xoviat:
Having Python installed in MSYS2 is a major red flag that will confuse the user. We might switch to MSYS2 and then just tell the user to run a command to uninstall Python. @ilayn What is that command?

dalito:
An update from my Openblas build. In the end I got three libraries in ..\build-openblas\OpenBLAS\
```
libopenblas_5f998ef_gcc7_2_0.dll (41361 KB)
libopenblas_5f998ef_gcc7_2_0.dll.a (5357 KB)
libopenblas_5f998ef_gcc7_2_0p-r0.2.20.a (49 993 KB)
```
The last file I renamed to "openblas.a".

xoviat:
@dalito Looking good!

andyfaff:
Really appreciate the work being put in to improving the documentation here. Lots of people trying to reproduce the windows buildchain from scratch is a vital effort.

ilayn:
OK, found what installs py2. It's the `mingw-.....-toolchain` packages. 

image

I tried to remove only that package but I got rejected due to dependencies. So we have to be extra careful explaining this, I think.

www/scipylib/building/windows.rst
```diff
@@ -32,185 +32,231 @@ information than this document will provide.
...
+    export BUILD_BITS=64
+
+Make sure that each variable makes sense. More specifically, make sure that the path that
+`OPENBLAS_ROOT` points to can be deleted. Make sure that the `OPENBLAS_COMMIT` points to
+the version that you want to build. Make sure that the architecture is correctly set to
+either 32 or 64 bit. And after you've made sure of that, build OpenBLAS.
+
+.. code:: shell
+
+    .\build_openblas.sh
```
ilayn: The slash is a forward one `./`

www/scipylib/building/windows.rst
```diff
@@ -32,185 +32,231 @@ information than this document will provide.
...
+.. code:: shell
+
+    export OPENBLAS_COMMIT=5f998ef
+    export OPENBLAS_ROOT="C:\\opt"
+    export BUILD_BITS=64
+
+Make sure that each variable makes sense. More specifically, make sure that the path that
+`OPENBLAS_ROOT` points to can be deleted. Make sure that the `OPENBLAS_COMMIT` points to
+the version that you want to build. Make sure that the architecture is correctly set to
+either 32 or 64 bit. And after you've made sure of that, build OpenBLAS.
```
ilayn: I would propose the following wording. 

```
Please check these variables' purpose for a moment. More specifically, make sure that you have 
write/delete access to the path `OPENBLAS_ROOT` points to. The output of the OpenBLAS build will 
be collected in this folder. Make sure that the `OPENBLAS_COMMIT` points to the correct OpenBLAS 
commit that you want to build in the cloned repo. In the future, `build_openblas` repository might get 
updated and you might want to get those updates by changing the commit. Make sure that the 
architecture is correctly set to either 32 or 64 bit. And after you've made sure of that, start 
the OpenBLAS build via 
```

www/scipylib/building/windows.rst
```diff
@@ -32,185 +32,231 @@ information than this document will provide.
...
+You don't necessarily need to build in that particular location, but it should be somewhere
+reasonable. Now clone the repository required to build OpenBLAS:
+
+.. code:: shell
+
+   git clone https://github.com/matthew-brett/build-openblas.git
+   cd build-openblas
+   git submodule update --init --recursive
+
+Now let's build OpenBLAS. To make sure that we're ready to build, type the following in
```
ilayn: I think it is better to perform this test before cloning the repository.

www/scipylib/building/windows.rst
```diff
@@ -32,185 +32,231 @@ information than this document will provide.
...
+is running. **Alternatively, if you'd rather download a pre-build OpenBLAS, download the**
+**one of the** `pre-built zip files`_ **and skip to the Installing OpenBLAS section below.**
+
+First, install MSYS2 using `these instructions`_. Make sure to install the correct
+architecture for the SciPy that you want to build (eg. 32 or 64 bit), and make sure to
+install the MinGW, and git packages. Open a MSYS2 terminal and type the following:
+
+.. code:: shell
+
+   pacman -S --needed base-devel
```
ilayn: I think splitting these for 32bit and 64 bit is better for one-liners. So I would propose:

```
First, install MSYS2 using `these instructions`_ including the `pacman` update instructions.
Occasionally during the updates the terminal might ask you to close the terminal but 
refuses to be closed and hang. If this happens you can kill it via Task Manager and continue 
with the instructions. Make sure to install the correct architecture for the SciPy that you 
want to build (eg. 32 or 64 bit). Now, you have three options for opening a terminal 
which are MSYS2, MINGW (32bit,64bit). After updating all the packages, now we are 
ready to install some more package bundles that we will need. Open a MSYS2 terminal 
and type the following depending on the architecture of your choice:

.. code:: shell

    pacman -S --needed base-devel mingw-w64-i686-toolchain mingw-w64-i686-cmake git

for 32-bit and 

.. code:: shell

    pacman -S --needed base-devel mingw-w64-x86_64-toolchain mingw-w64-x86_64-cmake git

for 64-bit build. It will prompt to whether install everything in these packages and 
you can simply accept all via hitting enter key at each step. 

We should be aware of the fact that these tools also install Python2, very similar to 
a virtual environment, which is only usable within MSYS2 terminals and we are 
**not** going to use it at any point. Therefore, after the OpenBLAS build we are 
done with MSYS2 and its terminals. We will not use for any step for building SciPy. 
Please keep this in mind to avoid nasty conflicts. 

You should now have all of the tools required to build OpenBLAS. Now change directories
to your favorite location with the following:
```

www/scipylib/building/windows.rst
```diff
@@ -32,185 +32,231 @@ information than this document will provide.
...
+   cd scipy
+   
+Now we need to copy the :code:`openblas.a` file that we've built earlier to the correct
+location. Find where Python is installed:
+
+.. code:: shell
+
+   python -c "import sys; print(sys.executable)"
+
+If your Python is installed somewhere like :code:`C:\\Program Files\\Python36\\python.exe`,
```
ilayn: Typical place for Win10 and Py3.6 is 

    C:\Users\<user name>\AppData\Local\Programs\Python\Python36\Lib

dalito:
Build successfully completed! The only problem was that the latest release of numpy (1.13.3) did not work. I had to update to the current development version.

Details of my setup:
 * Python-2.7.14 (from python.org) (I also have 3.4, 3.5, 3.6 on my system)
 * By default no Python version on path
 * MSYS2 
 * MinGW64
 * Microsoft Visual C++ 2008
 * [pypath.bat](https://gist.github.com/dalito/8356483ac2ffb2b8b468) on path (Script to activate one of the installed Python versions in "launcher-style")
 * numpy-1.14.0.dev0+77f9540

Steps after the openblas build:
 * Copied `openblas.a` to `C:\dev\Python27_64\libs`
 * Opened command prompt named "Visual C++ 2008 64-bit Command Prompt"
 * Activated python 2.7:
```
C:\dev\src\scipy>pypath -2
Added "C:\dev\Python27_64" and "C:\dev\Python27_64\scripts" to path.
```
 * Tested if gfortran is on path:
```
py-2|C:\dev\src\scipy>gfortran
gfortran.exe: fatal error: no input files
compilation terminated.
```
  - So gfortran was already on path. This is not starting the gfortran from MSYS2. To my surprise gfortran.exe is in "C:\dev\Python27_64\scripts" (which installer places this file in scripts dir?).
 * Started build with: "pip wheel -v -v -v ."
```
py-2|C:\dev\src\scipy>pip wheel -v -v -v .
(...waited more than an hour...)
done
  Stored in directory: c:\dev\src\whl_mine
Successfully built scipy
```
 * Installed the obtained wheel (scipy-1.1.0.dev0+unknown-cp27-cp27m-win_amd64.whl)
```
pip install -U --pre scipy
```
 * Instead of "python runtests.py --mode full" (which builds scipy once again!) the instruction should suggest to run the test with
```
py-2|C:\dev\src\scipy>python runtests.py -n
(...)
 1 failed, 12617 passed, 1430 skipped, 577 deselected, 112 xfailed, 7 xpassed in 1896.37 seconds
```
 *  Failing test was "test_weightedtau"

xoviat:
@pv I know your position, but so far no one has used MSYS2 to build SciPy. Not everyone may be as capable as you in this area and although this procedure takes longer if you can avoid any pitfalls, using MSYS2 to build SciPy poses a significant risk to beginners.

dalito:
I found out that I still had mingwpy (0.1.0b3) installed (that was giving me gfortran.exe under python/scripts) when I built scipy. - I will try once more after uninstalling mingwpy. I like @pv´s approach of building with MSYS2 since much less must be installed if it is doable.

xoviat:
@dalito Try it then and let me know how it goes. If it's an improvement, then we'll use that, but the biggest concern is that users will become confused about the MSYS2 python.

dalito:
After uninstalling mingwpy I could no longer build scipy from "Visual C++ 2008 64-bit Command Prompt" although I made gfortran from MSYS2 available (added `C:\dev\msys64\mingw64\bin` to path).

However, with MSYS2 building the scipy wheel was really simple. As @PV pointed out, the only important detail was to point to the MS-VC-compiled Python (in my case located in `C:\dev\python27_64`) to avoid use of the GCC-compiled Python within MSYS2):

1. Open "MSYS2 MinGW 64-bit" terminal for 64-bit Python (else "MSYS2 MinGW 32-bit")
2. Change to the directory with the scipy checkout:
`$ cd /c/dev/src/scipy`
3. Start building the wheel (3-times -v activates verbose reporting):
`"/c/dev/python27_64/python" -m pip wheel -v -v -v .`

This resulted in "scipy-1.1.0.dev0+59d4448-cp27-cp27m-win_amd64.whl" (in c:\dev\src\scipy). Result of running the test for this build:
`12618 passed, 1430 skipped, 577 deselected, 112 xfailed, 7 xpassed in 1885.37 seconds`

@ilayn maybe you want to try this route, too?

dalito:
@xoviat Do you also want to provide a link for the Microsoft Compiler for Python 2.7, [Microsoft Visual C++ Compiler for Python 2.7](https://www.microsoft.com/en-us/download/details.aspx?id=44266)?

xoviat:
I'll update the instructions for using MSYS2.

> Do you also want to provide a link for the Microsoft Compiler for Python 2.7,

Personally I'd rather not. I may be overridden on this, but making it easier to use Python 2.7 on windows is not a good thing. As I explained in scipy/scipy#8037, continuing to use MSVC 2008 is really sketchy and should be discouraged as much as possible.

ilayn:
I will try but I think this is going to be a bit harder for Py3 and MSVC 2017

xoviat:
@ilayn I don't think it actually will because Python finds MSVC 2017 automatically (with setuptools installed).

pv:
I should maybe say that the msys2 way is the way I've used the toolchain, for all Python versions.
setuptools will find the msvc compiler appropriate for each Python version, using MSVC shell or vcvarsall.bat is not needed.

xoviat:
Updated.

ilayn:
Well, before anything what I've noticed is that we have to install more stuff to get the Python3 working including Py3, `pip`, `setuptools`, `wheel` packages. 

As you can see from the following 

```
Ilhan Polat@CUDI MINGW64 /c/Users/Ilhan Polat/Documents/GitHub/scipy
$ python3 setup.py build
Traceback (most recent call last):
  File "setup.py", line 418, in <module>
    setup_package()
  File "setup.py", line 398, in setup_package
    from numpy.distutils.core import setup
ModuleNotFoundError: No module named 'numpy'

Ilhan Polat@CUDI MINGW64 /c/Users/Ilhan Polat/Documents/GitHub/scipy
$ cd ..

Ilhan Polat@CUDI MINGW64 /c/Users/Ilhan Polat/Documents/GitHub
$ cd numpy/

Ilhan Polat@CUDI MINGW64 /c/Users/Ilhan Polat/Documents/GitHub/numpy
$ pip3 install .
Processing c:/users/ilhan polat/documents/github/numpy
Installing collected packages: numpy
  Running setup.py install for numpy ... error
    Complete output from command C:/msys64/mingw64/bin/python3.exe -u -c "import setuptools, tokenize;__file__='C:/Users/ILHANP~1/AppData/Local/Temp/pip-l5d0brbi-build/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record C:/Users/ILHANP~1/AppData/Local/Temp/pip-1yt84eqv-record/install-record.txt --single-version-externally-managed --compile:

    Note: if you need reliable uninstall behavior, then install
    with pip instead of using `setup.py install`:

      - `pip install .`       (from a git repo or downloaded source
                               release)
      - `pip install numpy`   (last NumPy release on PyPi)


    Running from numpy source directory.
    Cythonizing sources
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
    ModuleNotFoundError: No module named 'Cython'
    Processing numpy/random/mtrand/randint_helpers.pxi.in
    Processing numpy/random/mtrand/mtrand.pyx
    Traceback (most recent call last):
      File "C:/Users/ILHANP~1/AppData/Local/Temp/pip-l5d0brbi-build/tools/cythonize.py", line 70, in process_pyx
        r = subprocess.call(['cython'] + flags + ["-o", tofile, fromfile])
      File "C:/msys64/mingw64/lib/python3.6\subprocess.py", line 267, in call
        with Popen(*popenargs, **kwargs) as p:
      File "C:/msys64/mingw64/lib/python3.6\subprocess.py", line 709, in __init__
        restore_signals, start_new_session)
      File "C:/msys64/mingw64/lib/python3.6\subprocess.py", line 997, in _execute_child
        startupinfo)
    FileNotFoundError: [WinError 2] The system cannot find the file specified

    During handling of the above exception, another exception occurred:

    Traceback (most recent call last):
      File "C:/Users/ILHANP~1/AppData/Local/Temp/pip-l5d0brbi-build/tools/cythonize.py", line 220, in <module>
        main()
      File "C:/Users/ILHANP~1/AppData/Local/Temp/pip-l5d0brbi-build/tools/cythonize.py", line 216, in main
        find_process_files(root_dir)
      File "C:/Users/ILHANP~1/AppData/Local/Temp/pip-l5d0brbi-build/tools/cythonize.py", line 208, in find_process_files
        process(cur_dir, fromfile, tofile, function, hash_db)
      File "C:/Users/ILHANP~1/AppData/Local/Temp/pip-l5d0brbi-build/tools/cythonize.py", line 167, in process
        processor_function(fromfile, tofile)
      File "C:/Users/ILHANP~1/AppData/Local/Temp/pip-l5d0brbi-build/tools/cythonize.py", line 81, in process_pyx
        raise Exception('Cython failed')
    Exception: Cython failed
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "C:/Users/ILHANP~1/AppData/Local/Temp/pip-l5d0brbi-build/setup.py", line 393, in <module>
        setup_package()
      File "C:/Users/ILHANP~1/AppData/Local/Temp/pip-l5d0brbi-build/setup.py", line 376, in setup_package
        generate_cython()
      File "C:/Users/ILHANP~1/AppData/Local/Temp/pip-l5d0brbi-build/setup.py", line 212, in generate_cython
        raise RuntimeError("Running cythonize failed!")
    RuntimeError: Running cythonize failed!

    ----------------------------------------
Command "C:/msys64/mingw64/bin/python3.exe -u -c "import setuptools, tokenize;__file__='C:/Users/ILHANP~1/AppData/Local/Temp/pip-l5d0brbi-build/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record C:/Users/ILHANP~1/AppData/Local/Temp/pip-1yt84eqv-record/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in C:/Users/ILHANP~1/AppData/Local/Temp/pip-l5d0brbi-build/

Ilhan Polat@CUDI MINGW64 /c/Users/Ilhan Polat/Documents/GitHub/numpy
$
``` 

So this is basically setting up a new Python3 machine. I am not sure if this path is easier.

xoviat:
@ilayn Have a look at the updated documentation. It requires you to specify the full path to Python.

ilayn:
@xoviat I don't follow why should it matter? I'm accessible to python3 as the errors show.

ilayn:
Failed command is `C:/msys64/mingw64/bin/python3.exe` which one should fail instead?

xoviat:
Probably:

`/c/Users/<user name>/AppData/Local/Programs/Python/Python36/python.exe`

ilayn:
Ah so I have to use native Python?

xoviat:
Yes.

ilayn:
So if I'm using native Python why am I doing it inside MSYS2 terminal which is foreign to almost all Win users?

xoviat:
`gfortran`

ilayn:
Wait I managed to do it outside MSYS2. Apparently that is not an answer.

xoviat:
Yes, but you had to install MinGW-builds.

xoviat:
There's only one option for building OpenBLAS: MSYS2. But building SciPy requires *a* gfortran. That can come from MSYS2 or MinGW-builds.

ilayn:
Yes, but why is that more complicated? I just clicked two more buttons and installed yet another program just like MSYS2. But anyways, that's another issue.  The build succeded by the way so probably we have to let the users know that it is the native Python that they should use.

xoviat:
> Yes, but why is that more complicated? 

I don't know. Some people think it is; some people think it isn't.

xoviat:
I think this is ready to merge?

dalito:
I could not yet build scipy on Python 3.6-64bit following the guide. I hit several problems:

* With German Visual Studio 2015/2017 the error messages are localised. When compiling scipy (or numpy), the compilation fails with a UnicodeDecodeError. The reason is that error messages from the compiler are localised and contain non-ascii characters which are not handled correctly (in subprocess).
 * Next I installed the Visual Studio 2017 Community Edition build tools as suggested in this PR and selected English as only language. 
 * With the compiler from VS2017 it was not possible to compile numpy 1.14.0.dev0, see numpy/numpy#9895
 * Then I changed to the older C/C++ compiler from Visual Studio 2015 which can be selected as an install option in the setup of the Visual Studio 2017 Community Edition build tools. Unfortunately, there is a [bug in vcvarsall.bat](https://developercommunity.visualstudio.com/content/problem/22031/unusable-vc-build-tools-2015-installed-with-vs2017.html) for the VS2015 compiler which had to be fixed (comment out line 8). After these changes numpy could be build with the VS2015 compiler (Version 19.00.24225.1 for x64).
 * After numpy was installed I could still not build scipy from MSYS2. The build failed with the following error: 
`fatal error C1047: The object or library file 'C:\\dev\\python36_64\\lib\\site-packages\\numpy\\core\\lib\npymath.lib' was created with an older compiler than other objects; rebuild old objects and libraries` The reason is that the VS2017 compiler was used which is selected by default if both VS2015 and VS2017 are installed (see [here](https://bugs.python.org/issue31340#msg301538)).
 * The next and hopefully final step will be to enforce using the VS2008 compiler also for the scipy build.
 
I would not recommend to install VS2017 at the moment for building scipy/numpy. Better stick with VS2015 (https://www.microsoft.com/en-us/download/details.aspx?id=53587).

xoviat:
@dalito To you have a trace for the UnicodeDecodeError? That should be fixed.

dalito:
No, unfortunately not. And I cannot easily recreate it since I removed the German VS 2015.

dalito:
I looked at the error in the debugger and it came from Python´s subprocess module and not from numpy or scipy.

xoviat:
Might be related to this:

https://github.com/numpy/numpy/blob/d21cecf4526fc463a063d058d92987633250f06c/numpy/distutils/exec_command.py#L252

xoviat:
What's the output of `python -c 'import locale; print locale.getpreferredencoding()'?

dalito:
"cp1252" (German Win 10 Pro)

xoviat:
Alright, now try this:

```
>> b'(some german character'.decode('cp1252')
```

dalito:
```
>>> b'äöüß'.decode('cp1252')
  File "<stdin>", line 1
SyntaxError: bytes can only contain ASCII literal characters.
```

ilayn:
I do have vc2017 installed by the way. Windows 10 machine.

dalito:
@ilayn - Which language(s) for VS2017 and for Win10?

ilayn:
@dalito Only C. In the installer I've chosen the boilerplate C++ toolchain option.

dalito:
I meant if you have any localization (German/Russian/English/...).

ilayn:
@dalito  Oh ahah, sorry, I need more coffee. Just english no extra localization.

pv:
@dalito: the reason why other people did not see the problem with compiling numpy, is that apparently you are trying to compile numpy so that it doesn't find the openblas installed.

pv:
If you try to enforce somehow a different VS compiler than what setuptools uses, then you will likely end up with binaries that have the wrong CRT.

dalito:
@xoviat A better download page for MS Build Tools may be: http://landinghub.visualstudio.com/visual-cpp-build-tools

dalito:
An update: On a different machine were I installed only VS2015 (non-localized), I could successfully build numpy, scipy and openblas (on Python-3.6.3-64bit, Win10Pro). 

The instruction are much better now. Thanks everyone! I think this PR can be merged now.

www/scipylib/building/windows.rst
```diff
@@ -32,185 +32,257 @@ information than this document will provide.
...
+.. code:: shell
+
+    pacman -S --needed base-devel mingw-w64-x86_64-toolchain mingw-w64-x86_64-cmake git
+
+for 64-bit build. It will prompt to whether install everything in these packages and 
+you can simply accept all via hitting enter key at each step. 
+
+We should be aware of the fact that these tools also install Python2, very similar to 
+a virtual environment, which is only usable within MSYS2 terminals and we are 
+**not** going to use it at any point. Therefore, after the OpenBLAS build we are
```
ilayn: Based on the recent changes, I think this sentence is no longer true as we keep on building scipy within MSYS2 right?

www/scipylib/building/windows.rst
```diff
@@ -32,185 +32,257 @@ information than this document will provide.
...
+Please keep this in mind to avoid nasty conflicts. 
+
+You should now have all of the tools required to build OpenBLAS. Now change directories
+to your favorite location with the following:
+ 
+.. code:: shell
+
+   cd /c/Users/MyUser/Downloads/
+   
+To make sure that we're ready to build, type the following in the terminal:
```
ilayn: I would add this sentence to get rid of the write access problem: 
```
If you already have a GitHub repository folder where you keep your own repos, it is better 
to use that location to keep things nice and tidy since we are going to clone yet another 
repository to obtain the source code, hence

.. code:: shell
    cd /c/<wherever the GitHub repo folder is>/GitHub
```

www/scipylib/building/windows.rst
```diff
@@ -32,185 +32,257 @@ information than this document will provide.
...
+
+   Copying the static library to /c/opt/64/lib
+
+Installing OpenBLAS
+===================
+
+If you see that line, then you might have OpenBLAS correctly, even if other failures
+occurred. Look in that folder for :code:`openblas.a`. If you find a file called something
+like :code:`libopenblas_5f998ef_gcc7_2_0.a`, just rename it to :code:`openblas.a` and continue.
+If the file isn't there, then poke around and try to find the file elsewhere. If you don't
```
ilayn: The file can't be arbitrarily placed right? If it is not in C:\opt, then `make .... install` has failed and the files didn't show up there. So this makes it a bit mysterious as if file can be placed somewhere that we have to figure out.

www/scipylib/building/windows.rst
```diff
@@ -32,185 +32,257 @@ information than this document will provide.
...
+.. code:: shell
+
+   cd C:\Users\MyUser\Downloads
+   git clone https://github.com/scipy/scipy.git
+   cd scipy
+   
+Now we need to copy the :code:`openblas.a` file that we've built earlier to the correct
+location. If your Python is installed somewhere like the following:
+
+..code:: shell
```
ilayn: A space after `..` needed to render

www/scipylib/building/windows.rst
```diff
@@ -32,185 +32,257 @@ information than this document will provide.
...
+
+   C:\Users\<user name>\AppData\Local\Programs\Python\Python36\python.exe
+
+
+Then you'll need to put the :code:`openblas.a` file somewhere like the following:
+
+..code:: shell
```
ilayn: Same space issue with this one

www/scipylib/building/windows.rst
```diff
@@ -32,185 +32,257 @@ information than this document will provide.
...
+
+If you see an error with the above command, :code:`gfortran` is not correctly installed.
+Go back to the "Building OpenBLAS" section and make sure that you have installed the correct
+tools.
+
+Now install the dependencies that we need to build and test SciPy. **It's important that you**
+**specify the full path to the native Python interpreter so that the built-in MSYS2 Python will**
+**not be used. Attempting to build with the MSYS2 Python will not work correctly.**
+
+**18 October 2017** A development NumPy is currently required to build. In addition, pip
```
ilayn: I would suggest mentioning the exact NumPy version via 

```
Until NumPy 1.14 is officially released, we have to work with the development version of the NumPy repository. You can still use the same OpenBLAS that you have built previously to also build NumPy. See the NumPy documentation....
```

dalito: It would be good to point to a specific location in the numpy docs. But as far as I can see the numpy documentation does also need an overhaul with respect to building on Windows like we do here.

xoviat: I'm also working on more upstream changes so that we can have a CRT-compatible fortran compiler on Windows. The goal is to be able to run `vcpkg install openblas && pip install https://github.com/numpy/numpy.git`. But that's for another day.

xoviat:
Changes addressed.

dalito:
@xoviat In the openblas section you only point out what would be wrong but don't tell what is correct. (lines 161ff)

Proposed new wording
"If you later observe build errors, verify that your build log reports that exactly one openblas library was found. "

```
FOUND:
   library_dirs = ['C:\opt\64\lib']
   libraries = ['openblas']
(...)
````

xoviat:
I've given everyone push access to my branch so that they can edit it directly. I personally don't want to spend more time on this.

dalito:
@xoviat - I don't plan to change more than this.

ilayn:
If anyone else doesn't have any other issues, I think this is ready to merge.

xoviat:
Ironically, we're about to have a much simpler build procedure for Python >= 3.5.

PR #242, opened 2018-01-20
DOC: update NumPy citation.

rgommers:
See "NumPy default citation" thread at
https://mail.python.org/pipermail/numpy-discussion/2017-September/077176.html


rgommers:
ping @stefanv, this was your proposal. we still should write that paper ... (after the scipy one)

rgommers:
Open for long enough and discussed on the mailing list, so merging.

stefanv:
That paper had a good run :)  https://scholar.google.com/scholar?cites=7372340385819983146&as_sdt=2005&sciodt=0,5&hl=en

So, while I did ask that we replace the citation, perhaps we should check with @teoliphant which book version he'd like cited, and whether there is an online DOI / URI for it?

rgommers:
Hmm yes, could make sense to cite the 2nd version. Travis?

Can't find a DOI/URI anywhere.

PR #244, opened 2018-01-22
pyemma replaced https with http

marscher:
sorry to bother you again @rgommers, but it turned out that our IT department can not serve SSL for external domains due to security rules.


rgommers:
no worries, thanks

PR #250, opened 2018-04-28
MAINT: remove build warnings, clean up outdated Sphinx extensions

rgommers:
There are only 2 build warnings left, those come from within the
Matplotlib extension so cannot be removed here.


charris:
I have no idea if this is correct, but we need to keep the site a little bit current :)

PR #256, opened 2018-07-18
Fix location of NumPy development docs redirects

stefanv:


www/.htaccess
```diff
@@ -14,13 +14,13 @@ RewriteRule ^(doc/) http://docs.scipy.org/ [R=permanent]
 RewriteRule ^(Download|Installing_SciPy) /install.html [R=permanent]
 RewriteRule ^(scipy_Example_List) https://docs.scipy.org/doc/scipy/reference/ [R=permanent]
 RewriteRule ^(Numpy_Example_List) https://docs.scipy.org/doc/numpy/reference/routines.html [R=permanent]
```
charris: This is rather strange, as this link is the same as it was below before replacement and seems to work OK. What is the difference?

pv:
Couldn't we instead redirect the `numpy-dev` to `devdocs`?
There may be other links pointing to it, since it used to work some time ago.

stefanv:
Feel free to push changes to my branch.

larsoner:
@mattip it looks like this could use a bit of work (not all examples converted, @pv's comment), could you push changes?

www/.htaccess
larsoner: @charris quoting @mattip from another thread where I asked a similar question (I think):

> NumPy is intending to host its own documentation going forward at numpy.org/doc

mattip:
I am not sure I understand this PR, when does this set of rules get used? Where is the MoinMoin site, and (for instance) where does it reference F2Py ?

pdebuyl:
Change by change:

-  `Numpy_Functions_by_Category` works via two redirects: to `https://docs.scipy.org/doc/numpy/reference/routines.html` then to the numpy website `https://numpy.org/doc/stable/reference/routines.html`. Could be shortened.
- `Tentative_NumPy_Tutorial` one redirect to `https://docs.scipy.org/doc/numpy-dev/user/quickstart.html` that is a 404.
- `F2Py` works via two redirects as well. Could be shortened to `https://numpy.org/doc/stable/f2py/` directly.
- `NumPy_for_Matlab_Users`: one redirect to `https://docs.scipy.org/doc/numpy-dev/user/numpy-for-matlab-users.html` that is a 404.

How much are these shortcuts used?

I would propose as a minimal solution to update the PR to fix `Numpy_Functions_by_Category` and `F2Py`. The other two are broken since probably two years, so could be dropped.

mattip:
Can you use intersphinx to ensure they are correct going forward?

PR #257, opened 2018-07-19
Add Code of Conduct to navigation sidebar

stefanv:
I asked a friend for feedback on our CoC, and she noted that it was not very easy to find.  I added it to the sidebar, and will make another PR to add it to the README as well.

### DIFF
--- a/www/_templates/sitenav.html
+++ b/www/_templates/sitenav.html
@@ -19,6 +19,7 @@ a.extlink:after {
       </ul>
       {% endif %}
     </li>
+    <a href="http://scipy.github.io/devdocs/dev/conduct/code_of_conduct.html">Code of Conduct</a></li>
     <li {% if pagename == 'install' -%}class="active"{%- endif -%}>
       <a href="{{ pathto('install') }}">Install</a></li>
     <li {% if pagename == 'getting-started' -%}class="active"{%- endif -%}>
--- a/www/_templates/sitenav.html
+++ b/www/_templates/sitenav.html
@@ -10,6 +10,7 @@ a.extlink:after {
 {% if not pagename.startswith('scipylib/') -%}
 <div class="well sidebar-nav">
   <ul class="nav nav-list">
+
     <li {% if pagename == 'about' -%}class="active"{%- endif -%}>
       <a href="{{ pathto('about') }}">About SciPy</a>
       {% if pagename in ('about', 'stackspec') -%}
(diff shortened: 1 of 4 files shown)


rgommers:
+1 good idea

rgommers:
It's in the wrong place though, it should be in the sidebar of https://scipy.org/scipylib/index.html. That's the front page for SciPy the library; the CoC is not for the ecosystem.

stefanv:
So... hrm. SciPy.org seems like the landing place for all SciPy related 
activities? E.g., there are links at the top (the big icons!) to file bug 
reports to the library. The CoC would be most discoverable here, but if we 
only mention it elsewhere we should probably also remove items related to 
the library from this page, other than perhaps a pointer to the library?

pv:
The bug report links lead to bug trackers of all projects, not only scipy.

pv:
and ditto for all other links, iirc thetes nothing project såecific on the front page

rgommers:
@stefanv yes this is partly the problem with scipy.org - it's for the ecosystem, but it's too limited for that, while the library pages are hidden. that's one of the things I wanted to discuss next week.

rgommers:
We could address that by a new page "Code of Conducts" which links to all the ones for individual projects. That keeps it very visible.

rgommers:
- SciPy CoC: http://scipy.github.io/devdocs/dev/conduct/code_of_conduct.html
- Pandas CoC: https://github.com/pandas-dev/pandas-governance/blob/master/code-of-conduct.md
- NumPy CoC: doesn't exist (yet)
- Matplotlib CoC: doesn't exist (yet)
- Sympy CoC: https://github.com/sympy/sympy/blob/master/CODE_OF_CONDUCT.md
- Jupyter/IPython CoC: https://github.com/jupyter/governance/blob/master/conduct/code_of_conduct.md

stefanv:
I updated the PR with this suggestion (CoC page added that lists CoC's for all 6 projects shown on the website, and an entry for the SciPy library).

rgommers:
I'll wait for a day or two with merging, but this looks good to me.

rgommers:
No objections, so in it goes. Thanks Stefan!

rgommers:
@nellev FYI https://scipy.org/codes-of-conduct.html - would be good not to forget to update this in case Matplotlib gains itself a CoC.

NelleV:
Yep, I'll keep this in mind!

tacaswell:
For several years we have had "Matplotlib is a welcoming, inclusive project, and we try to follow the Python Software Foundation Code of Conduct in everything we do."  on the main matplotlib.org page.

rgommers:
@tacaswell do you want to put that back up your website for the time being? Then we can link to / mention that.

tacaswell:
It is still there, but we have not dedicated CoC page.

PR #262, opened 2018-08-09
deprecate using homebrew to build numpy, scipy

mattip:
As noted in numpy/numpy#11454 we should not recommend using homebrew to build scientific python packages.

### DIFF
--- a/www/install.rst
+++ b/www/install.rst
@@ -84,12 +84,18 @@ For Python 3.5 with `Macports <https://www.macports.org>`_ execute this command 
 
 .. rubric:: Homebrew
 
-Homebrew is not the correct way to install NumPy. We reccomend
-installing python, aliasing it, then using `pip`::
+Homebrew no longer packages NumPy or other packages of the scientific Python
+stack. If you're a Homebrew user, install
+[Python3 with Homebrew](https://docs.brew.sh/Homebrew-and-Python)::
 
     brew install python
-    # alias python to python3
-    python -m pip install numpy scipy ...
+    brew link python
+    python -c"import sys; print(sys.version)"
+    # if this does not show python3, these steps failed
+
+and the rest with pip::
+
+    python -m pip install numpy scipy matplotlib
 
 .. _individual-packages:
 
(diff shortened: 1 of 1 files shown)


www/install.rst
```diff
@@ -84,9 +84,12 @@ For Python 3.5 with `Macports <https://www.macports.org>`_ execute this command
 
 .. rubric:: Homebrew
 
-You can install NumPy, SciPy, and Matplotlib, with::
+Homebrew is not the correct way to install NumPy. We reccomend
```
rgommers: I suggest: "Homebrew no longer packages NumPy or other packages of the scientific Python stack. If you're a Homebrew user, installing only Python with Homebrew and the rest with Pip is the recommended method::"

www/install.rst
```diff
@@ -84,9 +84,12 @@ For Python 3.5 with `Macports <https://www.macports.org>`_ execute this command
...
+installing python, aliasing it, then using `pip`::
 
-    brew tap homebrew/science && brew install python numpy scipy matplotlib
+    brew install python
+    # alias python to python3
```
rgommers: can you turn this into a working command?

www/install.rst
```diff
@@ -84,9 +84,12 @@ For Python 3.5 with `Macports <https://www.macports.org>`_ execute this command
...
+    python -m pip install numpy scipy ...
```
rgommers: This should be a working command as well: `python -m pip install numpy scipy matplotlib`

mattip:
Reworked in the direction you requested. I do not have a Mac, I just gleaned information.

www/install.rst
```diff
@@ -84,9 +84,18 @@ For Python 3.5 with `Macports <https://www.macports.org>`_ execute this command
...
 .. rubric:: Homebrew
 
-You can install NumPy, SciPy, and Matplotlib, with::
+Homebrew no longer packages NumPy or other packages of the scientific Python
+stack. If you're a Homebrew user, install
+[Python3 with Homebrew](https://docs.brew.sh/Homebrew-and-Python)::
 
-    brew tap homebrew/science && brew install python numpy scipy matplotlib
+    brew install python
+    brew link python
```
rgommers: I think that should be `brew unlink python && brew link python`

www/install.rst
```diff
@@ -84,9 +84,18 @@ For Python 3.5 with `Macports <https://www.macports.org>`_ execute this command
...
+    python -c"import sys; print(sys.version)"
```
rgommers: typo, needs space after `-c`

rgommers:
Merged, thanks Matti

PR #266, opened 2018-09-01
Fix broken Anaconda link

caichinger:
The section [Scientific Python distributions](https://scipy.org/install.html#scientific-python-distributions) lista Anaconda. However, the link is at the moment pointing to
> https://www.continuum.io/downloads (<-- dead)

whereas it should in my opinion point to

> https://www.anaconda.com/download/ (<-- new)

I replaced the link in this PR.

### DIFF
--- a/www/install.rst
+++ b/www/install.rst
@@ -12,7 +12,7 @@ For many users, especially on Windows, the easiest way to begin is to
 download one of these Python distributions, which include all the key
 packages:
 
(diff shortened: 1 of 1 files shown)


rgommers:
thanks @caichinger

rgommers:
fixed a second instance of this broken link in 19385f9

caichinger:
thanks for merging @rgommers (I was not aware of the second one, next time I will `grep` :)

PR #283, opened 2019-02-18
MAINT: a number of edits to FAQ questions.

rgommers:
Closes gh-https://github.com/scipy/scipy/issues/9825

### DIFF
--- a/www/scipylib/faq.rst
+++ b/www/scipylib/faq.rst
@@ -247,6 +247,17 @@ to do so is 1.2.x.
 The first release of NumPy to support Python 3.x was NumPy 1.5.0.
 Python 3 support in SciPy was introduced in 0.9.0.
 
+Does NumPy/SciPy work with PyPy?
+################################
+
+In general, yes. Recent improvements in PyPy_ have made the scientific Python
+stack work with PyPy. The NumPy and SciPy projects run PyPy in continuous
+integration and aim to further improve support over time.
+Since much of NumPy and SciPy is implemented as C extension
+modules, the code may not run any faster (for most cases it's significantly
+slower still, however PyPy is actively working on improving this). As always
+when benchmarking, your experience is the best guide.
+
 Does NumPy/SciPy work with Jython or C#/.NET?
 #############################################
 
@@ -514,3 +525,4 @@ everyone time.
 .. _Bokeh: https://bokeh.pydata.org/en/latest
 .. _Altair: https://altair-viz.github.io
 .. _Plotly: https://plot.ly
+.. _PyPy: http://pypy.org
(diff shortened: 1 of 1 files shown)


www/scipylib/faq.rst
```diff
@@ -238,29 +239,25 @@ matrix argument for solving `generalized eigenvalue problems`_.
...
-(CPython) interpreter.
+No, neither are supported. Jython never worked, because it runs on top of the
+Java Virtual Machine and has no way to interface with extensions written in C
+for the standard Python (CPython) interpreter.
 
-Does NumPy/SciPy work with IronPython (.NET)?
-#############################################
+Some years ago there was an effort to make NumPy and SciPy compatible with .NET.
+Some users at the time reported success in using NumPy with `Ironclad
+<https://code.google.com/archive/p/ironclad>`_ on 32-bit Windows.
```
mattip: Could we add a Does NumPy/SciPy work with PyPy? with the answer "In general, yes. Since much of NumPy and SciPy is implemented as c-extension modules, the code may not run any faster, but as always when benchmarking, your experience is the best guide."

www/scipylib/faq.rst
```diff
@@ -500,13 +499,18 @@ and a scalar:
...
-.. _Enthought: https://www.enthought.com/
-.. _Matplotlib: http://matplotlib.org/
+.. _f2py: http://www.f2py.com
+.. _Enthought: https://www.enthought.com
+.. _Anaconda: https://www.anaconda.com
+.. _Quansight: https://www.quansight.com
+.. _Matplotlib: http://matplotlib.org
+.. _Bokeh: https://bokeh.pydata.org/en/latest
+.. _Altair: https://altair-viz.github.io
+.. _Plotly: https://plot.ly
```
mattip: The canonical link for PyPy is https://www.pypy.org, not sure how to link it in to my comment above

www/scipylib/faq.rst
rgommers: yes good question. I will tweak the answer a bit, because it now looks like it will be as fast or faster, but in general it's way slower and not yet recommended for production usage

rgommers: added a FAQ entry, please review. I didn't literally say "not yet ready for production" because that hopefully will change soon and we shouldn't discourage people too much, but some sort of disclaimer is needed.

www/scipylib/faq.rst
rgommers: added, thanks

www/scipylib/faq.rst
mattip: Looks appropriate for now, thanks

rgommers:
okay, straightforward cleanups, so merging this. thanks for the review @mattip

PR #284, opened 2019-02-21
Add back NumPy citation as secondary source

stefanv:
I discussed this change with @rgommers.  I asked at some point that we remove the citation to my and Gaël's paper in favor of Travis's book.  However, I've received feedback that a) the new source is not available openly online, b) is harder to cite (no DOI, URL, etc.), and c) sends academic credit to a place where it is not very useful.  So, this PR adds the paper back as a *secondary* reference.  I leave the decision of whether to incorporate this change fully up to the team.

I think the solution, in the longer term, is to write a new NumPy paper.  Perhaps to accompany the release of NumPy 2.0, which shouldn't be too far off.  It will document all the changes that happened over the past two years, which are numerous.  I know there is interest in writing such a paper, and it may be the first to give some historical perspective on NumPy.

Also note that there is no "how to cite NumPy" section on numpy.org.  We may want to remedy that.

### DIFF
--- a/www/citing.rst
+++ b/www/citing.rst
@@ -68,12 +68,21 @@ __ http://scitation.aip.org/content/aip/journal/cise/9/3/10.1109/MCSE.2007.58
 __ https://doi.org/10.1109/MCSE.2011.36
 __ http://scitation.aip.org/content/aip/journal/cise/13/2/10.1109/MCSE.2011.36
 
+
 NumPy
 #####
 
-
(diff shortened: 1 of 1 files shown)


rgommers:
Thanks Stefan, LGTM. Let's leave it open for a few days to give people time to comment.

And +1 for a new paper, even if there's no NumPy 2.0:)

mattip:
merge?

rgommers:
Merged, thanks @stefanv

Issue #99, opened 2015-07-11
Improving topical software


pv:
The topical software page is a direct translation of a wiki page. 

- It is very long, and not very easy to read.
- It is not regularly updated --- authors of some projects add their packages there, but this does not seem to occur very commonly.
- It is far from comprehensive, not listing e.g. stuff on pypi

There are probably better ways to do it, potentially including
(i) splitting the long list to one metadata file per entry,
(ii) replacing sections by hierarchical tags,
(iii) automatically importing data from pypi (whole Science/Research category), using the local metadata for tagging.

PR #287, opened 2019-03-27
Update topical-software.rst

hvparks:
add metrolopy to science: basic tools


rgommers:
Thanks @hvparks. I'm happy to add MetroloPy, however I'm thinking we should make a small subsection on units and also add other popular packages. Probably this is the list of most relevant ones: `pint`, `unyt`, `astropy.units`, `python-quantities` and `quantiphy`. What do you think?

hvparks:
> I'm thinking we should make a small subsection on units and also add other popular packages

That would be great.  You might consider making the section also cover packages that do uncertainty propagation, e.g. `uncertainties`, `gtc`, and `scinum`.  Its not exactly the same thing, but if you are dealing with a value that has a unit it is likely to also have an uncertainty.

hvparks:
> section also cover packages that do uncertainty propagation

Though I guess you do already have a section on automatic differentiation

rgommers:
Before I forget this completely I'll push the green button and then look at the restructure I proposed. Thanks @hvparks

rgommers:
>  and then look at the restructure I proposed.

Actually, I'll leave that for now, in favor of a more structural solution (see gh-99).

PR #288, opened 2019-04-17
BLD: remove matplotlib.sphinxext.only_directive

rgommers:
This was remove in MPL 3.0, and we didn't seem to use it anyway.

Can now use builtin ``only:: html`` rather than ``htmlonly::``


rgommers:
This works, just deployed a new version. Is trivial, so merging.

PR #290, opened 2019-04-27
Update sponsor/donation content

rgommers:
As discussed on the mailing list: https://mail.python.org/pipermail/scipy-dev/2019-April/023457.html

- remove "sponsored by Enthought" from banner
- update SciPy library donation page
- add NumFOCUS to front page as fiscal sponsor

This also needs https://github.com/scipy/scipy-sphinx-theme/pull/11. And docs.scipy.org needs an update as well. Will update that once the scipy-sphinx-theme PR is merged.

Note that this just changes the essentials. A larger overhaul is still desirable.

PR #293, opened 2019-05-08
Update donate button 

rgommers:
- links to the correct Salsalabs page that NumFOCUS maintains for SciPy
- changes the visual from yellow to blue

also fix a link that @WarrenWeckesser pointed out in gh-290


rgommers:
Note that this is really a minimal fix to get the right links in. If anyone with web design skills wants to make that button more appealing, and/or add more content, then please do!

PR #299, opened 2019-06-29
GH: add issue template

pv:
Attempt to direct people to use project bug trackers instead of this
one for errors in documentation.


ilayn:
Simple enough to warrant a direct merge. Thanks @pv

PR #304, opened 2019-08-08
reference numpy.org for all things numpy

mattip:
Part of numpy/numpy.org#25 and friends

### DIFF
--- a/www/docs.rst
+++ b/www/docs.rst
@@ -3,12 +3,12 @@ Documentation
 
 Documentation for core SciPy Stack projects:
 
-* `Numpy <http://docs.scipy.org/doc/numpy/>`__
+* `Numpy <http://www.numpy.org/doc>`__
 * `Scipy <http://docs.scipy.org/doc/scipy/reference/>`__
-* `Matplotlib <http://matplotlib.org/contents.html>`_ 
+* `Matplotlib <http://matplotlib.org/contents.html>`_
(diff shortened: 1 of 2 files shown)


www/getting-started.rst
```diff
@@ -69,7 +69,7 @@ is to take a look at the following online resources:
 
 - `Python Scientific Lecture Notes <http://www.scipy-lectures.org/index.html>`_
 
-- `NumPy User Guide <http://docs.scipy.org/doc/numpy-dev/user/>`_ 
+- `NumPy User Guide <http://www.numpy.org/devdoc/user/>`_
```
rgommers: typo, should be `devdocs` not `devdoc`

www/docs.rst
```diff
@@ -3,7 +3,7 @@ Documentation
 
 Documentation for core SciPy Stack projects:
 
-* `Numpy <http://docs.scipy.org/doc/numpy/>`__
+* `Numpy <http://www.numpy.org>`__
```
rgommers: https://numpy.org/doc/

www/getting-started.rst
mattip: fixed

www/docs.rst
mattip: fixed

mattip:
It seems #256 also would be good to review and merge as part of this change

larsoner:
Comments addressed and LGTM, thanks @mattip

PR #305, opened 2019-08-09
Announce SciPy 1.3.1

tylerjereddy:
Announce SciPy `1.3.1`, if we're still doing that stuff here -- can't keep track of all the website changes!


rgommers:
yes we are, thanks Tyler

PR #307, opened 2019-08-17
Revised Install Page

albi3ro:
I've just reworked the content on the install page, hopefully making it a bit clearer and smoother.

### DIFF
--- a/www/install.rst
+++ b/www/install.rst
@@ -1,113 +1,99 @@
 ===================
-Installing packages
+Installation
 ===================
 
-These are general instructions for installing packages in the SciPy
-ecosystem.
+Installations methods include:
 
-Scientific Python distributions
--------------------------------
+* :ref:`Distributions <distributions>`
+* :ref:`Pip <pip-install>`
+* :ref:`Package Manager <package_manager>`
+* :ref:`Source <source>`
+* :ref:`Binaries <binaries>`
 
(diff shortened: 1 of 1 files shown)


rgommers:
Thanks @albi3ro, looks great from a quick browse! Will review in detail during the weekend.

www/install.rst
```diff
@@ -1,113 +1,99 @@
...
+Installation
 ===================
 
-These are general instructions for installing packages in the SciPy
-ecosystem.
-
-Scientific Python distributions
--------------------------------
-
-For many users, especially on Windows, the easiest way to begin is to
```
rgommers: I'd like to keep some form of this recommendation. Perhaps at the end of the new section, in italics. Something like _"If you don't know which installation method you need or prefer, we recommend you use a scientific Python distribution (see next section)"_

albi3ro: Added that, but made it more specific to Anaconda, since "scientific Python distribution" itself still leaves open confusing choices.

rgommers: Sounds good, thanks!

rgommers:
Merged, thanks @albi3ro. 

This reorganizes the existing content mostly, for easier navigation. Some parts are a bit outdated and can probably be removed (e.g. the binaries section; also the `--user` needs a warning that it's _not_ the right thing to do with virtualenvs or conda envs). Let's leave that for a follow-up.

PR #309, opened 2019-08-23
Two links go to location for scipy tutorials and docs

albi3ro:
The route to get from scipy.org to the tutorials and documentation was confusing and involved unnecessary intermediaries.  I cleaned up two links.

### DIFF
--- a/www/_templates/sitenav.html
+++ b/www/_templates/sitenav.html
@@ -53,7 +53,7 @@ a.extlink:after {
   <ul class="nav nav-list">
     <li class="nav-header">Core packages:</li>
     <li><a class="extlink" href="http://www.numpy.org">Numpy</a> 
-    <li><a class="extlink" href="{{ pathto('scipylib/index') }}">SciPy library</a>
+    <li><a class="extlink" href="https://docs.scipy.org/doc/scipy/reference/">SciPy library</a>
     <li><a class="extlink" href="http://matplotlib.org">Matplotlib</a> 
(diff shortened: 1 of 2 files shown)


rgommers:
This changes the meaning of that link from going to the "home page" to going to the documentation. So that doesn't seem quite right to me. The page that it currently points to is admittedly bare-bones, but there's a number of things in the sidebar of https://scipy.org/scipylib/index.html that are not in the docs (e.g. license, report bug, donations).

The fastest way to get to the docs from scipy.org is `Documentation > SciPy`.

Perhaps the confusing thing here is that https://scipy.org is about "SciPy the ecosystem" rather than "SciPy the library". That's what we would really like to fix, then this becomes a lot more straightforward. So far I had in mind to have a replacement "ecosystem site" before changing scipy.org, however we could go without that. There really isn't that much content on the current site that we'd have to remove, and some of the things that are there now (like SciPy Stack) have become irrelevant. We could easily say that `scipy.org` will be about SciPy the library, while keeping a page like https://scipy.org/about.html that explains the ecosystem context.

You got me thinking @albi3ro, thanks:) Does the above all make sense? If so, we can propose it on the `scipy-dev` mailing list.

rgommers:
I'll close this PR, since I'm fairly sure this hides content that's useful. I will bring this up on the mailing list now for feedback.

rgommers:
Email sent: https://mail.python.org/pipermail/scipy-dev/2019-September/023731.html

PR #317, opened 2019-10-22
MAINT: add SciPy arXiv preprint

tylerjereddy:
- ~still a draft, formatting probably isn't quite right yet~
- ~not sure if we want to remove the old/website citation~
- may want to simply link to the BibTeX copy/paste website instead of using
all that inline space for the example entry
- ~I need to add the title of the preprint--forgot that!~

The main motivation is that it looks like, informally, the SciPy 1.0 manuscript will likely get published around January 2020, so we may want to try to get some benefit/citations on the preprint until that time.

### DIFF
--- a/www/citing.rst
+++ b/www/citing.rst
@@ -13,6 +13,53 @@ __ http://scitation.aip.org/content/aip/journal/cise/13/2
 SciPy (the library)
 ###################
 
+There is now a preprint available for citing usage of SciPy until
+the journal article is available:
+
+* Pauli Virtanen, Ralf Gommers, Travis E. Oliphant, Matt Haberland,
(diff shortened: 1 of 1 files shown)


tylerjereddy:
I think I've addressed most of my original concerns now--that latest html build of the citation page on my system looks like this for this feature branch:

image

I removed the old website citation--hopefully that's not controversial.

@mdhaber any concerns/ requests?

rgommers:
> may want to simply link to the BibTeX copy/paste website instead of using all that inline space for the example entry

We can change later if desired, let's get this up now. Not strange to treat SciPy a bit more extensively than other packages, it's `scipy.org` after all.

rgommers:
Rendering was broken due to leaving in a html tag, fixed up in commit d53600f

PR #319, opened 2019-11-10
Move some news entries to past-news, reduce length of main page

rgommers:


rgommers:
This is simple housekeeping, merging straight away

PR #321, opened 2019-11-13
DOC: survey update

mkg33:
I removed the outdated ad about taking part in our survey and added news about the results.


rgommers:
Thanks Maja!

PR #327, opened 2019-12-06
Announce NumPy 1.18.0rc1 release

charris:


rgommers:
It's happened much more often, but it's news that I typically clean up after a while cause it's not very relevant in the long term. So consolidating rc1/rc2/final to a single 1.xx.0 release

rgommers:
deployed now

PR #344, opened 2020-05-02
Add scikit-network in machine learning and AI

tbonald:


rgommers:
Merged, thanks @tbonald!

tbonald:
Great!
Is there a way to be referenced in the scikit list?

http://scikits.appspot.com/scikits

The package is on PyPI:
https://pypi.org/project/scikit-network/

Thanks.

pv:
In principle it should appear there automatically, but it seems the site is broken now. E.g. this shouldn't be a blank page: https://scikits.appspot.com/scikit-data

PR #376, opened 2020-12-03
Announce NumPy 1.20.0rc1.

charris:


rgommers:
Thanks Chuck!

mattip:
Whohoo!! 

Do we announce it on numpy.org as well?

rgommers:
Only the final release I'd say

PR #386, opened 2021-03-25
Announce SciPy 1.6.2

tylerjereddy:
Announce SciPy `1.6.2`

I'll likely do the docs build/upload another day though.


rgommers (approved):
Thanks @tylerjereddy. 

Docs build later makes sense, I agree we don't need to wait for that.

PR #389, opened 2021-04-26
Announce SciPy 1.6.3

tylerjereddy:
Announce SciPy 1.6.3

### DIFF
--- a/www/index.rst
+++ b/www/index.rst
@@ -215,6 +215,9 @@ News
 .. role:: news-date
    :class: news-date
 
+SciPy 1.6.3 released :news-date:`2021-04-25`
+    See :doc:`/scipylib/download`.
+
 NumPy 1.20.2 released :news-date:`2021-03-27`
     See :doc:`/scipylib/download`.
 


tupui (approved):

tupui:
Thanks Tyler. Looking at other similar PRs LGTM so merging.

rgommers:
Thanks for another useful release shipped Tyler!

Website updated now. @tupui it's better not to merge unless you can also deploy to the server that hosts scipy.org and docs.scipy.org. When we redo the website we should migrate away from that server to something saner (GitHub Pages or Netlify), but for now it's only a few people who have ssh access.

tupui:
> Website updated now. @tupui it's better not to merge unless you can also deploy to the server that hosts scipy.org and docs.scipy.org. When we redo the website we should migrate away from that server to something saner (GitHub Pages or Netlify), but for now it's only a few people who have ssh access.

Oh sorry I didn't know 🙇‍♂️  
Not sure how/when the new website will happen. Are we going (have we already) to make some open calls for that?

rgommers:
No worries at all. It's a bit outdated unfortunately.

> Not sure how/when the new website will happen. Are we going (have we already) to make some open calls for that?

I'd like to wait till we know about a possible CZI cycle 4 grant, that would give us someone with bandwidth to coordinate a complete redesign at once. If that's not successful, we can tackle it in a fashion that's more incremental.

PR #425, opened 2021-11-08
Use the new structure for News

tupui:
News are now automatically generated from the `/news` folder. Each file is a news. The way it works is still being defined.

Right now it will create a single news which will be displayed on the front page. It will (after a theme update) link to the current news listing we already have.

- https://github.com/scientific-python/scientific-python-hugo-theme/pull/66

To be defined:

- Do we want to separate releases from news?
- Do we want to create a news per release? This is needed right now.
- Other requirements?

EDIT:

Upstream changes made it possible to stay with a single page and have the banner defined in the preamble. This PR update the theme and add the preamble keyword.

- https://github.com/scientific-python/scientific-python-hugo-theme/pull/71

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -45,11 +45,6 @@ languages:
         buttonlink: "/install"
         # Hero image (from static/images/___)
         image: logo.svg
-      # Customizable navbar. For a dropdown, add a "sublinks" list.
-      news:
-        title: SciPy 1.7.2 released
-        content: 2021-11-05
-        url: /news
 
       keyfeatures:
         features:
--- a/content/en/news.md
+++ b/content/en/news.md
@@ -1,6 +1,8 @@
 ---
 title: News
 sidebar: false
+newsHeader: SciPy 1.7.2 released!
+date: 2021-11-05
 ---
 
 ### SciPy 1.7.2 released


rgommers:
> * Do we want to separate releases from news?

I think it's fine to mix them. Although I don't think bugfix releases are all news-paragraph worthy.

> * Do we want to create a news per release? This is needed right now.

We should for minor releases.

> * Other requirements?

not that I can think of.

tupui:
Thanks Ralf. Then if we mix releases with other news, is it ok that on the front page we could have an entry which is not a release entry?

The way it works:

```
en/news
    |_ news.md
    |_ news_link.md
```
The latest news is listed in `news_link.md` and it directly links to `news.md`. This is to not break the current news page here and in NumPy.

Now if you want to add another news with some content which would not be in `news.md`, it will add this standalone page at the url `/news`. So if we have a release on the front page and get redirected to `news.md`, then we need to add in here a link to `/news`. Or we add a news tab somewhere.

Second option is to not use the redirection and then clicking on the release news would get to `/news`. Here we would need to click on another link to be directed to `news.md`. The issue is that as we will add standalone news pages, `news.md` will get pushed at the bottom of the page (or we add another rule to fix the position of some pages...).

cc @jarrodmillman @stefanv

rgommers:
> Then if we mix releases with other news, is it ok that on the front page we could have an entry which is not a release entry?

Sure why not. If a release is say 5 months old, why wouldn't we replace it with something more recent/relevant?

I don't understand the rest of your comment. Why would we need separate files for non-release news? All we need is a single entry for the bar on the front page, and a news page - just like on https://numpy.org/news/

tupui:
> Sure why not. If a release is say 5 months old, why wouldn't we replace it with something more recent/relevant?

Ok great! I was not sure if we wanted to always show releases info here or not.

> I don't understand the rest of your comment. Why would we need separate files for non-release news? All we need is a single entry for the bar on the front page, and a news page - just like on https://numpy.org/news/

This is if at some point we would want to add news with more content like a blog type of entry. We could have guest posts to showcase SciPy usages, decision stories, things about a new maintainers, etc. Such content would not leave well on this single page.

rgommers:
> This is if at some point we would want to add news with more content like blog-type entries. We could have guest posts to showcase SciPy usage, decision stories, things about a new maintainer, etc. Such content would not live well on this single page.

The question is if that should be on this site, or on a separate blog for example. One consideration is that we may also want to add translations, and adding a lot of long-form content kind of clashes with that.

tupui:
> The question is if that should be on this site, or on a separate blog for example. One consideration is that we may also want to add translations, and adding a lot of long-form content kind of clashes with that.

Indeed. Ok but then if we say we never want to have long content and only use a single page, then I suppose it is better to have the url `scipy.org/news`. So we need to revert some changes on the theme to allow that. Right now it would be `scipy.org/news/news` (or anything else but just `/news`).

rgommers:
> So we need to revert some changes on the theme to allow that

https://numpy.org/news/ already exists and shouldn't be broken, so those changes don't sound good to me anyway.

stefanv:
I think we can make both schemes work quite easily (single file vs category for multiple posts).

I suggest adding the news content into /news/index.md. That makes news a page bundle that renders the way it does right now.

With single file, the title of the advertised news item will have to be manually specified, as it is now. We could make it slightly more convenient by having it in the index.md preamble (say `newsBanner` or similar). That way, you don't ever need to touch the config file.

If anything, Hugo has too much flexibility, so we just need to experiment with a few designs and see what feels good for everyone.

stefanv:
We don't necessarily need the folder: we could go with news.md instead. From the Hugo perspective they are more or less equivalent.

_index files are used to provide some text before a category listing (blog posts, news items, etc.).

index.md files turn a directory into a page bundle: more or less equivalent to a single file, like news.md, except that it can contain its own images and other media.

tupui:
> We don't necessarily need the folder: we could go with news.md instead. From the Hugo perspective they are more or less equivalent.

Great then if we can use the preamble to specify the banner that would be great. (FYI, my last push uses your PR to render as it uses the `newsLink` to redirect to the banner. I used this to see how the site would look.)

stefanv:
See https://github.com/scientific-python/scientific-python-hugo-theme/pull/71

tupui:
> See [scientific-python/scientific-python-hugo-theme#71](https://github.com/scientific-python/scientific-python-hugo-theme/pull/71)

Thanks @stefanv for the quick update of the theme! Now all looks good and this is ready.

stefanv (approved):

jarrodmillman:
Do we want to keep
```
      news:
        title: SciPy 1.7.2 released
        content: 2021-11-05
        url: /news
```
in `config.yaml`?

tupui:
> Do we want to keep
> 
> ```
>       news:
>         title: SciPy 1.7.2 released
>         content: 2021-11-05
>         url: /news
> ```
> 
> in `config.yaml`?

I don't think so. I find it more convenient to have this in the preamble of the single file. This way there is just a single file to take care of.

jarrodmillman:
I removed the old config cruft in e220fcf.

tupui:
Thanks @jarrodmillman. I thought I did but with all the changes seems I forgot 😅Good catch!

jarrodmillman (approved):

PR #428, opened 2021-11-24
Announce SciPy 1.7.3

tylerjereddy:
* Announce SciPy 1.7.3

* I think the `news` stuff has already been reorganized a bit since I last did this so let me know if there's a better way

### DIFF
--- a/content/en/news.md
+++ b/content/en/news.md
@@ -1,9 +1,13 @@
 ---
 title: News
 sidebar: false
-newsHeader: SciPy 1.7.2 released!
-date: 2021-11-05
+newsHeader: SciPy 1.7.3 released!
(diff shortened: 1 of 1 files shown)


tupui (changes requested):
Thanks @tylerjereddy. It's just missing at the end the link to the release notes. Sorry it's far in the file to make a suggestion. Otherwise you did it right 😃

tylerjereddy:
I added a link to the release notes; I don't see the connection to forward-porting though--it seems that these links just point to the github release page, which is already available.

tupui (approved):
Thanks for the quick update! You're totally right with the links. For some reason I thought it linked to the doc instead of GH. All is good then 😃

PR #430, opened 2021-12-30
Rename master to main

jarrodmillman:
This is for after @rgommers renames the master branch main.


rgommers (approved):
This LGTM, seems complete. This PR is for when the main `scipy` repo moves from `master` to `main`. 

For moving this repo itself, I think we need to mess with the Netlify config - which is you @jarrodmillman?

PR #434, opened 2022-01-06
DOC: Refactor about section

tupui:
Refactor the about us section to follow NumPy's choices (html and logos are from NumPy):

* Move donation page to about
* Add partners
* Add sponsors
* Add CI providers (not sure here about who to add and in which order if that matters) and JetBrains in acknowledgement section
* Add steering council members


tupui:
I am adding the steering council members right now. I will clean on the main repo the governance page. Regarding the members, I don't know if there are emeritus or anything. @rgommers I will let you edit this. Also shall I add the BDFL and release manager sections there?

jarrodmillman:
You may be able to find the old steering committee members on web archive or in the repo history for scipy.org.  For example, around 2010, it included me, Eric Jones, Robert Kern, Travis Oliphant, and Stefan van der Walt:
https://web.archive.org/web/20100612192251/http://scipy.org/Developer_Zone

tupui:
Thanks @jarrodmillman, I will start an emeritus section then. Maybe we should have somewhere a log for things like that?? Something simple saying, on this date xxx join or left this position in the org.

jarrodmillman:
I believe the committee was unofficial for years, but first listed on the website in Feb 2008:
https://web.archive.org/web/20080221234022/http://www.scipy.org:80/Developer_Zone

Stefan joined around August 2008:
https://web.archive.org/web/20080827214948/http://www.scipy.org/Developer_Zone

I stopped contributing around 2012 (when I started grad school) and it looks like the committee was removed from the website in 2013:  https://github.com/scipy/scipy.org/pull/10

I am not sure when the more recent committee was formed.

rgommers:
@tupui I think you should move all the people who moved to emeritus maintainer status to emeritus here as well.

We've never really used this steering committee, it was basically all maintainers active for >1 yr. Similar for the very old one - Eric Jones was long gone by then, and in 2010 it was mostly Pauli and me as maintainers. So start/end year probably doesn't make sense, since we never maintained that list of people.

> Also shall I add the BDFL and release manager sections there?

That seems like a good idea to me. They're important roles. Maybe just indicate behind the names, rather than new sections on the page?

tupui:
Ok thanks I will do the changes!

jarrodmillman:
I agree we don't need dates, but it isn't true that the old steering committee wasn't used.  We regularly met, organized sprints, made roadmaps, and other things at least in the early years.  As I got involved in other things, I think we stopped meeting and planning.

One big difference from the current committee is that we used to be the steering committee for NumPy, SciPy, and the SciPy conference.  Another is that we had no formal documents, so how people joined and how they were removed was unclear.

rgommers:
Thanks for the historical context @jarrodmillman, always interesting!

content/en/about.md
```diff
@@ -8,6 +8,43 @@ and wider scientific Python community. For more information on our governance
...
+
+* Anne Archibald
+* Eric Jones
+* Eric Moore
+* Eric Quintero
+* Jaime Fernández del Río
+* Jarrod Millman
+* Robert Kern
+* Stéfan van der Walt
+* Travis Oliphant (Creator of SciPy)
```
stefanv, line 46: SciPy was created when Travis (multipack), Eric (Cluster of Workers, GA, and weave), and Pearu (f2py) joined efforts.

This parenthetical may be a bit out of place in this list, because not all founders' names appear here. Perhaps rather mention it in a paragraph about the history of the library?

jarrodmillman: According to [github](https://github.com/scipy/scipy/graphs/contributors?from=2001-01-28&to=2004-06-03&type=c) between Jan 28, 2001 – Jun 3, 2004, Travis, Pearu, and Eric are the only contributors.  That may miss SVN accounts that didn't translate to github.  It also doesn't track the contributions to the code that were made before they were combined.

rgommers: I suggest to list all three as co-creators

content/en/about.md
```diff
@@ -8,6 +8,43 @@ and wider scientific Python community. For more information on our governance
...
+* Paul van Mulbregt
+* Pauli Virtanen (BDFL)
+* Ralf Gommers (Chair)
+* Tyler Reddy (Release manager)
+* Warren Weckesser
+
+Emeritus:
+
+* Anne Archibald
+* Eric Jones
```
tupui, line 39:
```suggestion
* Eric Jones (co-creator of Scipy)
```

content/en/about.md
tupui, line 46:
```suggestion
* Travis Oliphant (co-creator of SciPy)
```

content/en/about.md
```diff
@@ -8,6 +8,43 @@ and wider scientific Python community. For more information on our governance
...
+* Eric Moore
+* Eric Quintero
+* Jaime Fernández del Río
+* Jarrod Millman
+* Robert Kern
```
tupui, line 44:
```suggestion
* Jarrod Millman
* Pearu Peterson (co-creator of SciPy)
* Robert Kern
```

jarrodmillman (approved):

content/en/about.md
```diff
@@ -8,6 +8,44 @@ and wider scientific Python community. For more information on our governance
...
+* Paul van Mulbregt
+* Pauli Virtanen (BDFL)
+* Ralf Gommers (Chair)
+* Tyler Reddy (Release manager)
+* Warren Weckesser
+
+Emeritus:
+
+* Anne Archibald
+* Eric Jones (co-creator of Scipy)
```
stefanv, line 39:
```suggestion
* Eric Jones (co-creator of SciPy)
```

stefanv (approved):

layouts/shortcodes/partners.html
```diff
@@ -0,0 +1,34 @@
+<div class="partner-images">
+    <a href="https://www.quansight.com">
+        <img src="/images/logos/quansight.svg" width="200" alt="Logo of Quansight" />
```
rgommers, line 3: This is the old logo. I don't have the new Quansight one at hand, but for now can you use this Quansight Labs one instead?

https://github.com/Quansight-Labs/quansight-labs-site/blob/main/files/images/QuansightLabs_logo_V2.png

layouts/shortcodes/sponsors.html
```diff
@@ -0,0 +1,40 @@
+<div class="sponsor-images">
+    <a href="https://www.chanzuckerberg.com">
+        <img src="/images/logos/CZI.svg" alt="Logo of the Chan Zuckerberg Initiative" />
+    </a>
+    <a href="https://www.tidelift.com">
+        <img src="/images/logos/tidelift.png" width="200" alt="Logo of Tidelift" />
+    </a>
+    <a href="https://numfocus.org">
+        <img src="/images/logos/NumFOCUS_sponsored_project_logo.svg" width="200" alt="Logo of NumFOCUS" />
```
rgommers, line 9: NumFOCUS itself does not give us enough direct funding to list it here. I'd rather do it the same way as done on https://numpy.org/about/

rgommers (commented):
Looks good modulo two small things. Thanks @tupui!

layouts/shortcodes/sponsors.html
tupui: Ok I thought the SDG grants and the additional donations would be enough to have them here.

rgommers: Donations aren't coming from NumFOCUS, they just play the role of fiscal host. So with SDG's we're getting max $15k/yr if we get a grant every round. At least according to the numpy guidelines, that's below the bar to be listed.

rgommers (approved):
This is a significant improvement, let's get it in. Thanks @tupui & reviewers!

PR #435, opened 2022-01-09
MAINT: Analytics using plausible.io with scientific-python

### LINKED ISSUE #338: Add Google Analytics?
One way of measuring impact and which functions are used frequently would be via website traffic. I'd be interested in setting up Google Analytics in order to collect that data. Based on [this](https://www.similarweb.com/website/scipy.org#overview) estimate of the number of hits scipy.org gets per month (~3 million), we wouldn't exceed the [traffic limit](https://support.google.com/analytics/answer/1070983?hl=en) for Google Analytics free version (10 million hit/month). [Setup](https://support.google.com/analytics/answer/1008015?hl=en) just requires adding the same snippet of code to the `<head>` of every page we want to track.

1. Shall I give this a shot?
2. Would the right place to add the code to this repo be an [`extrahead` block](https://www.sphinx-doc.org/en/master/templating.html) in [`layout.html`](https://github.com/scipy/scipy.org/blob/master/www/_templates/layout.html)?
2. Should this be added to scipy-sphinx-theme, too? Would adding the code directly to the `<head>` section of its [`layout.html`](https://github.com/scipy/scipy-sphinx-theme/blob/master/_theme/scipy/layout.html) make the code appear on every page?

tupui:
Closes #338

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -20,6 +20,9 @@ params:
   font:
     name: "Lato"
     sizes: [400,900]
+  plausible:
+    dataDomain: scipy.org
+    javaScript: "https://analytics.scientific-python.org/js/plausible.js"
 
 languages:
   en:
--- a/config.yaml
+++ b/config.yaml
@@ -22,7 +22,6 @@ params:
     sizes: [400,900]
   plausible:
     dataDomain: scipy.org
-    javaScript: "https://analytics.scientific-python.org/js/plausible.js"
 
 languages:
   en:
--- a/README.md
+++ b/README.md
@@ -33,3 +33,10 @@ with `read:org` permissions. The token has to be exported as `GH_TOKEN`.
 export GH_TOKEN=xxxxxxxxxx
 make teams
 ```
+
+## Analytics
+
+The service [Plausible.io](https://plausible.io) is used to gather simple
+and privacy-friendly analytics for the site. The dashboard can be accessed
+[here](https://analytics.scientific-python.org/scipy.org).
+The Scientific-Python community is managing the account.


config.yaml
```diff
@@ -20,6 +20,9 @@ params:
   font:
     name: "Lato"
     sizes: [400,900]
+  plausible:
+    dataDomain: scipy.org
+    javaScript: "https://analytics.scientific-python.org/js/plausible.js"
```
stefanv, line 25: You don't need to specify javaScript here, since that is the default.  But it won't hurt either.

stefanv:
Note that you'd also need to pin the latest version of the theme for it to work.

config.yaml
tupui, line 25:
```suggestion
```

tupui: Let's remove it then since either way it's linked to the theme.

tupui (commented):
> Note that you'd also need to pin the latest version of the theme for it to work.

I updated the theme, did I miss something else?

tupui:
How does it work to see the metrics?? Do you have a dashboard specific to SciPy then? Is this public or restricted to some people/org/teams?

stefanv (approved):

stefanv:
> How does it work to see the metrics?? Do you have a dashboard specific to SciPy then? Is this public or restricted to some people/org/teams?

I will send you credentials for the SciPy dashboard. Credentials are limited to a single website, and AFAIK you can have as many users as you like.

stefanv:
Account invitation sent to you and Ralf. You will be able to add others.

tupui:
Thank you Stéfan 🙏 @rgommers is this ok for you?

tupui:
The dashboard looks great! Lot of options and we could even make some data public if we wanted to.

tupui:
I added some basic note in the readme to say that there is a dashboard.

rgommers (approved):
LGTM, thanks @tupui and @stefanv

rgommers:
@stefanv could you please resend the invite to my personal email?

tupui:
For me it was in my spam.

mdhaber:
I'd appreciate an invitation, too! Thanks for setting this up, all.

tupui:
Just sent it to you @mdhaber on your personal email.

PR #437, opened 2022-01-22
Update theme

jarrodmillman:
It was getting annoying depending on the fresh theme, since it doesn't get updated regularly.  So we copied the missing pieces over to the scientific python theme.  Nothing should be different.   We just want to get this rolled out before we start updating things.


tupui (approved):
LGTM, thanks @jarrodmillman

PR #440, opened 2022-01-25
DOC/MAINT: refactor community and contribute pages

### LINKED ISSUE #426: Update community
Following #421, the mailing list only page will serve as the community page.

We can copy the content from NumPy to start.

tupui:
Closes #426

I used content from NumPy's website.


config.yaml
```diff
@@ -101,14 +101,14 @@ languages:
...
-            - text: Mailing Lists
-              link: /mailing-lists
+            - text: Community
+              link: /community
             - text: SciPy Conference
               link: https://conference.scipy.org/
+            - text: Contribute
+              link: /contribute
+            - text: Code of Conduct
+              link: http://scipy.github.io/devdocs/dev/conduct/code_of_conduct.html
```
melissawm, line 111: Is there a reason why this is not under https://scipy.org?

tupui: AFAIK there is no reason, if we say it's better suited here than in the main doc I could move it here (or have the doc in both locations)

content/en/contribute.md
```diff
@@ -0,0 +1,102 @@
...
+We can barely begin to list the contributions a graphic designer can make here.
+Our docs are parched for illustration; our growing website craves images --
+opportunities abound.
+
+
+### Fundraising
+
+SciPy was all-volunteer for many years, but as its importance grew it became
+clear that to ensure stability and growth we'd need financial support. [This
+SciPy'19 talk](https://www.youtube.com/watch?v=dBTJD_FDVjU) explains how much
```
melissawm, line 98: I don't know if this talk is what you wanted here :)

tupui: Although it's more focused on NumPy, I think it explains well why funding is important. But I could also remove this sentence if you think it's not appropriate (or adjust the timestamp to link to the end of the presentation).

That's something we could have on the scientific-python website itself. cc @jarrodmillman

config.yaml
melissawm: Got it - i didn't realize this is how the site was structured!

content/en/contribute.md
jarrodmillman: Adding the timestamp would probably be helpful.  We should also continue this conversation over the next year.  We definitely will work on this as part of the Scientific Python project.  I will follow up soonish.

config.yaml
stefanv: Maybe in another PR we can copy the text onto the website to have it render there as its own page.

stefanv:
This looks good to me, so I'm going to merge for now, and then we can iterate as necessary.  Thanks @tupui!

PR #442, opened 2022-01-27
Run linkcheck in CI

jarrodmillman:
@tupui Would you double-check this?  In particular, take a look at f8ff69c2b037b.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -32,3 +32,6 @@
 [[redirects]]
   from = "/scipylib/faq.html"
   to = "/faq/"
+
+[[plugins]]
+  package = "netlify-plugin-checklinks"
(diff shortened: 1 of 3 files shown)


netlify.toml
```diff
@@ -23,7 +23,7 @@
 
 [[redirects]]
   from = "/scipylib/dev-zone.html"
-  to = "/dev-zone/"
+  to = "/community/"
```
tupui, line 26: I think this should be
```suggestion
  to = "/contribute/"
```

tupui (commented):
Thanks @jarrodmillman! Good catch. These plugins are nice 👍

tupui (approved):
LGTM, thanks @jarrodmillman

PR #449, opened 2022-02-03
update theme

jarrodmillman:
Various cleanups and one user-visible change.  The shortcuts/sidebar has been fixed.  It now doesn't scroll up with the rest of the page and the links work.  It also makes the appearance more like what you see with pydata-sphinx-theme.

## Old

screenshot

## New

screenshot


tupui (approved):
Very nice improvement of the sidebar. Thanks!

I did have time to comment on the theme side, I noticed there is a slight offset of the highlight. What I mean is that the extra small vertical line is not exactly centred with the text. Also it could be in the middle of the big line.

PR #465, opened 2022-03-09
Broken Link to Governance

NaincyKumariKnoldus:
Hey @noatamir and @tupui I have fixed this. And I loved to do this. Please review my PR.
Thanks!

#464

PR #466, opened 2022-03-09
Update theme

jarrodmillman:


tupui:
Was this a glitch before? Now it's again above 0.9.

jarrodmillman:
It varies here and I've been having to manually restart it on netlify to get it to not complain for the last few PRs.  Normally it is around .89 to .91.  So the .9 threshold is a little too high.  I initially thought about lowering it to .88, but I noticed that in #465 it was .86.  So I decided to go with .85 for now.  I've been keeping an eye on it and some upcoming theme changes should improve performance a bit.  So I suspect we should be able to increase it to closer to .9 soon.

tupui:
Ok thanks then let's go with that.

PR #472, opened 2022-04-19
Update Package Names for Fedora

abe-101:
Following the installation guide to install scipy system wide on fedora:
`sudo dnf install numpy scipy python-matplotlib ipython python-pandas sympy python-nose atlas-devel`
Produces the error:
```
No match for argument: scipy
No match for argument: sympy
Error: Unable to find a match: scipy sympy
```
I've updated the guide to include the correct package names:
`sudo dnf install python-numpy python-scipy python-matplotlib ipython python-pandas python-sympy python-nose atlas-devel`

### DIFF
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -90,7 +90,7 @@ using apt-get:
 
 using dnf:
 
-    sudo dnf install numpy scipy python-matplotlib ipython python-pandas sympy python-nose atlas-devel
+    sudo dnf install python-numpy python-scipy python-matplotlib ipython python-pandas python-sympy python-nose atlas-devel
 
 ## Mac
 
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -86,7 +86,7 @@ using apt-get:
 
     sudo apt-get install python-numpy python-scipy python-matplotlib ipython ipython-notebook python-pandas python-sympy python-nose
 
-## Fedora 22 and later
+## Fedora
 
 using dnf:
 
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -90,7 +90,7 @@ using apt-get:
 
 using dnf:
 
-    sudo dnf install python-numpy python-scipy python-matplotlib ipython python-pandas python-sympy python-nose atlas-devel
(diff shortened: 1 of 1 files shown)


tupui (approved):
Thank you @abe-101. LGTM but I am not a Fedora user.

@jarrodmillman I think you use Fedora right? Anything else to do here?

tupui:
Here https://developer.fedoraproject.org/tech/languages/python/scipy.html I can see that they use `python3-...` instead. Not sure what's preferred.

abe-101:
> Here https://developer.fedoraproject.org/tech/languages/python/scipy.html I can see that they use `python3-...` instead. Not sure what's preferred.
> 
> Also I thought we cleaned this page a bit more than that. We are still referring to other packages than SciPy. cc @melissawm

It seems that `python-...` will point to the version of python the user is running. I'm not sure what the preferred way is.
Based on a chat I had with other fedora users it seems `python3-...` would be "more correct".
I've gone ahead and updated it.
Thank you @tupui for pointing it out.

jarrodmillman:
I would use `python3-ipython` as well.  We don't need `python3-nose` or `atlas-devel`.

The whole page needs some attention, but that can be in a separate PR.

content/en/install.md
```diff
@@ -90,7 +90,7 @@ using apt-get:
 
 using dnf:
 
-    sudo dnf install numpy scipy python-matplotlib ipython python-pandas sympy python-nose atlas-devel
+    sudo dnf install python3-numpy python3-scipy python3-matplotlib ipython python3-pandas python3-sympy python3-nose atlas-devel
```
jarrodmillman, line 93:
```suggestion
    sudo dnf install python3-numpy python3-scipy python3-matplotlib python3-ipython python3-pandas python3-sympy python3-pytest
```

jarrodmillman (changes requested):

jarrodmillman:
I would also change the heading from `Fedora 22 and later` to `Fedora`.

jarrodmillman (approved):

jarrodmillman:
Thanks!

PR #476, opened 2022-05-06
Fix a dead link for contributors in the about page.

AtsushiSakai:
I fixed a dead link for contributors in the [about](https://scipy.org/about/) page.
The THANKS.txt file does not exist now.
I think GitHub’s [contributors](https://github.com/scipy/scipy/graphs/contributors) page is a good option.
Let me know if there is any other good option.


tupui (approved):
Thank you @AtsushiSakai. It's a good replacement.

PR #479, opened 2022-05-12
Update theme

jarrodmillman:
@tupui Would you mind carefully comparing:
- https://scipy.org/
- https://deploy-preview-479--scipy-org.netlify.app/

I think it should be the same. But I am starting to remove a lot of the old theme stuff to make it easier to maintain and modify the theme, so it would be great if you could take a close look before we merge this.


tupui (commented):
I don't see any differences 😃

PR #484, opened 2022-06-08
Fix broken link to build instructions

dymil:
I checked with [archive.org](https://web.archive.org/web/20210303001538/https://scipy.github.io/devdocs/building/index.html) that this is the right page to link to.

### DIFF
--- a/content/en/download.md
+++ b/content/en/download.md
@@ -47,7 +47,7 @@ To check out the latest **SciPy** sources:
 
 Build instructions for SciPy can be found in its documentation. The
 latest version can be found at:
-<https://scipy.github.io/devdocs/building/index.html>
+<https://docs.scipy.org/doc/scipy/building>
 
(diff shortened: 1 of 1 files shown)


tupui (commented):
Thank you @dymil for the PR. I am afraid this is not correct. The new link you are suggesting is pointing at the latest released documentation while the current link is pointing to the development version. This is intended as build instructions (and for that matter all development instructions) evolve quickly and it's better to point to the current version on `main`.

dymil:
Hi @tupui ,
My apologies. Unless I'm missing something though, it remains the case that the current link is broken (produces a 404 error). Would [this link](https://scipy.github.io/devdocs/dev/contributor/building.html) into the devdocs be the proper destination?

tupui:
Oh yes indeed, apologies. Yes this would be the correct link.

PR #492, opened 2022-07-23
Add twitter to socials in footer

tupui:
Maybe we can also add Slack, Discord (but discord would be the Scientific Python one) and discuss? WDYT?

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -85,7 +85,7 @@ languages:
         - link: https://github.com/scipy/scipy
           icon: github
         - link: https://twitter.com/SciPy_team
-        icon: twitter
+           icon: twitter
       quicklinks:
         column1:
           title: ""
--- a/config.yaml
+++ b/config.yaml
@@ -84,6 +84,8 @@ languages:
       socialmedia:
         - link: https://github.com/scipy/scipy
           icon: github
+        - link: https://twitter.com/SciPy_team
+        icon: twitter
       quicklinks:
         column1:
           title: ""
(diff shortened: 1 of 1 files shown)


config.yaml
```diff
@@ -84,6 +84,8 @@ languages:
       socialmedia:
         - link: https://github.com/scipy/scipy
           icon: github
+        - link: https://twitter.com/SciPy_team
+        icon: twitter
```
tupui, line 88:
```suggestion
           icon: twitter
```

config.yaml
```diff
@@ -84,6 +84,8 @@ languages:
       socialmedia:
         - link: https://github.com/scipy/scipy
           icon: github
+        - link: https://twitter.com/SciPy_team
+           icon: twitter
```
tupui, line 88:
```suggestion
          icon: twitter
```

rgommers (approved):
Thanks @tupui, Twitter icon is nice to have.

> Maybe we can also add Slack, Discord (but discord would be the Scientific Python one) and discuss? WDYT?

Slack is a private channel so there's nowhere to link to. Discord isn't part of the SciPy project; now that after all these years the site is about SciPy the project and nothing else, I'd prefer to keep it that way. So an icon doesn't seem to fit. If we do want to link out to a representative set if relevant community sites/projects, we should probably add a section under Community? It can also point to the Ecosystem part of numpy.org (best to leave it there and not repeat that kind of info on scipy.org)

tupui:
Ok yes I will think about adding a section on the community page.

PR #495, opened 2022-08-22
Make install page specific to SciPy and link to build docs

rgommers:
This is a quick cleanup in response to https://github.com/scipy/scipy/issues/16308#issuecomment-1221684861, which pointed out that details on building from source were missing on the install page.


tupui (approved):
Good cleanup thanks Ralf.

PR #503, opened 2022-10-19
Update team members

tupui:
FYI, @stefanv you are still not in the GH list for the web team and I think you should be. I manually update the list to have you there.


stefanv:
Won't this just get overwritten next time you generate the gallery?

stefanv:
For now, I've added myself to the web team to simplify your life. Do you need any other people added to teams?

tupui:
haha thanks. I think otherwise it's good. At leat with what I know :)

tupui:
Pushing this now.

PR #509, opened 2023-01-20
Add LANL institutional partner

tylerjereddy:
* pleased to report I'm covered at 20 % time and effort by LANL for SciPy/NumPy work now; this is not the same pot of funding as NASA ROSES, though those grant meetings still make sense as a coordination point I think

* permission to use the LANL logo is tricky--I think this is good enough unless they tell me otherwise; I couldn't quite decide if we wanted the list of partners and then the subset of logos below, but since it is just Quansight logo for now I just went under that logo, feel free to adjust..

### DIFF
--- a/content/en/about.md
+++ b/content/en/about.md
@@ -73,6 +73,8 @@ Partners include:
 
 {{< partners >}}
 
+- Los Alamos National Laboratory (Tyler Reddy)
+
 ## Donate
 
 SciPy will always be 100% open source software, free for all to use and


tupui:
Awesome! I will look next week

rgommers (approved):
LGTM. And congrats Tyler, that's good news!

tupui (approved):
LGTM too. Let's merge.

If you want to add LANL's logo, you need to update this https://github.com/scipy/scipy.org/blob/main/layouts/shortcodes/partners.html
and add the logo there https://github.com/scipy/scipy.org/tree/main/static/images/logos

PR #514, opened 2023-04-20
fix broken links to release notes

j-bowhay:
0.19.x links were broken. Also placed in chronological order


tupui (approved):
Good catch, thanks Jake 😃

PR #526, opened 2023-06-28
Announce SciPy 1.11.1

tylerjereddy:
Announce SciPy `1.11.1`, which was released rapidly to account for a licensing violation in `1.11.0`. The latter release was yanked from PyPI just now.

I did not add any information about the yanking here, but at least it should be clear enough on PyPI.

### DIFF
--- a/content/en/news.md
+++ b/content/en/news.md
@@ -1,9 +1,13 @@
 ---
 title: News
 sidebar: false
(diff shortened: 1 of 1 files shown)


tupui (approved):
LGTM, thank you Tyler for handling all of that 🙇

ilayn:
Indeed, thanks for the lightning fast response!

PR #539, opened 2023-11-07
Update theme (0.4)

jarrodmillman:
This is a major update. It notably adds a dark/light theme and integrates a lot of styling from pydata-sphinx-theme. This will improve accessibility and should make the experience cleaner for users moving between the homepage and the documentation.

See https://github.com/scientific-python/scientific-python-hugo-theme/blob/main/CHANGELOG.md

Currently, you are (essentially) using v0.1 of the theme.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -50,3 +50,5 @@
 
 [[plugins]]
   package = "netlify-plugin-checklinks"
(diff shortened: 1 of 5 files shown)


tupui (approved):
LGTM, thanks Jarrod!

PR #558, opened 2024-02-20
Move keyfeatures to a grid of cards

jarrodmillman:
Building on functionality from the v0.11 release. The keyfeature panel for the homepage will be removed in a future theme release. This is part of our large effort to adopt styling from pydata-sphinx-theme where possible. It also makes the frontpage behave more like the other pages.

You can see more examples of cards (in a grid) here: 
- https://theme.scientific-python.org/shortcodes/#grid

This will also make it easier to modify your project homepage without having to override `layouts/index.html`.

The styling is a little different than before due to adopting pydata-sphinx-theme styling. But that styling is being used on many community sites and is being reviewed for accessibility.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -49,21 +49,6 @@ languages:
         # Hero image (from static/images/___)
         image: logo.svg
 
-      keyfeatures:
-        features:
-          - title: Fundamental algorithms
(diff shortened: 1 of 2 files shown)


tupui (approved):
Thanks @jarrodmillman all good and it looks nice too 😃

PR #567, opened 2024-03-11
Use grid of cards for team gallery

jarrodmillman:
We will need to change grid with grid1 when we upgrade the theme (i.e., we just need to delete one character).


lucascolley:
The formatting is slightly off for small avatar sizes:

image

PR #572, opened 2024-03-26
Update Teams page

rgommers:
This is the output of `make teams`:
- adds Sean Cheah to the Maintainers team (Cc @thalassemia for visibility)
- updates an avatar that changed since the last run


lucascolley (approved):
LGTM!

image

PR #580, opened 2024-05-22
Replace Salsa with OpenCollective link

### LINKED ISSUE #579: Donate button redirects to random tech site
I think @tupui is aware of this problem, that is same as https://github.com/astropy/astropy/issues/16480 . I see that you indeed have the same problem as of 2024-05-22 over at https://scipy.org/about/#donate . FYI

tupui:
Closes #579

Also update theme.


tupui:
Merging this now as the current state is clearly an issue for us.

PR #590, opened 2024-09-21
Add a DOI to the default citation

RichardLitt:
There is no DOI link in the default citation, although there is one in the Bibtex. Adding it here makes it more likely for people to use it. 

This follows the format that Numpy uses, [here](https://numpy.org/citing-numpy/).


lucascolley (approved):
LGTM, thanks @RichardLitt !

PR #592, opened 2024-10-02
update teams

lucascolley:


lucascolley:
@tylerjereddy I think you also have permission to remove Dietrich from the Triage Team? Then we could update this PR and put it in.

tylerjereddy (commented):
I've removed Dietrich Brunn from the Triage team on the GitHub SciPy org as requested. They are already listed on maintainers team, so I think you're good to go with an update to mirror that here.

lucascolley:
thanks Tyler!

PR #595, opened 2024-12-16
Update installation page

lucascolley:
- Modernise discussion around installing from PyPI or conda-forge
- Include `uv` and `pixi` as options
- Add explanation of how to install with static type stubs
- remove mention of Python distributions and Anaconda `defaults` (https://github.com/scipy/scipy.org/pull/595#discussion_r1887539401)

cc @rgommers @stefanv @jarrodmillman @tupui @jorenham

### DIFF
--- a/content/en/beginner-install.md
+++ b/content/en/beginner-install.md
@@ -45,10 +45,10 @@ Other options include:
   and the IEP interactive development environment; Supports Linux,
   Windows, and Mac.
 
+{{< admonition note >}}
 Users in large, non-university institutions may want to read Anaconda's
-helpful blog on ["when is Anaconda free to use?"][anaconda-blog]
-
-[anaconda-blog]: https://www.anaconda.com/blog/update-on-anacondas-terms-of-service-for-academia-and-research
+helpful blog on ["when is Anaconda free to use?"](https://www.anaconda.com/blog/update-on-anacondas-terms-of-service-for-academia-and-research)
+{{< /admonition >}}
 
 After installing a scientific Python distribution,
 see next steps in [the SciPy user guide][scipy-user-guide].
--- a/content/en/beginner-install.md
+++ b/content/en/beginner-install.md
@@ -1,0 +1,70 @@
+---
+title: Beginner Installation Guide
+sidebar: false
+---
+
+{{< admonition tip >}}
+This is the beginner installation guide.
+If you are comfortable with using a terminal and happy to learn
+how to use a package manager, check out 
+[the main installation guide](./install.md)!
+{{< /admonition >}}
+
+- [JupyterLite](#jupyterlite)
+- [Scientific Python Distributions](#distributions)
+- [Installing globally with `pip`](#pip-global)
+
+<a name="jupyterlite"></a>
+
+## JupyterLite
+
+To try out SciPy, you don't even need to install it!
+You can use SciPy in your browser at https://jupyter.org/try-jupyter/lab/ -
+just write `import scipy` in one of the notebook "cells" and hit play.
+
+For next steps, see [the SciPy user guide][scipy-user-guide].
+
+<a name="distributions"></a>
+
+## Scientific Python Distributions
+
+Python distributions provide the language itself, along with the most
+commonly used packages and tools. These downloadable files require
+little configuration, work on almost all setups, and provide all the
+most commonly used scientific Python tools.
+[Anaconda](https://www.anaconda.com/download/) works on Windows, Mac,
+and Linux, and is best suited to beginning users.
+Other options include:
+- [WinPython](https://winpython.github.io): Another free distribution
+  including scientific packages and the Spyder IDE; Windows only, but
+  more actively maintained and supports the latest Python 3 versions.
+- [Pyzo](https://pyzo.org): A free distribution based on Anaconda
+  and the IEP interactive development environment; Supports Linux,
+  Windows, and Mac.
+
+Users in large, non-university institutions may want to read Anaconda's
+helpful blog on ["when is Anaconda free to use?"][anaconda-blog]
+
+[anaconda-blog]: https://www.anaconda.com/blog/update-on-anacondas-terms-of-service-for-academia-and-research
+
+After installing a scientific Python distribution,
+see next steps in [the SciPy user guide][scipy-user-guide].
+
+<a name="pip-global"></a>
+
+## Installing globally with `pip`
+
+If you already have Python installed, you can install SciPy globally
+with `pip` by executing the following in a terminal/shell:
+
+    python -m pip install scipy
+
+You may see this recommended in tutorials or classes, but it is not
+recommended to install packages globally on your system.
+The recommended way to install SciPy with `pip` is to use a virtual
+environment - see [Installing with `pip`](./install.md#installing-with-pip).
+
+{{< admonition note >}}
+For more information on why this is not a recommended installation method,
+read about [virtual environments in the Python Packaging User Guide](https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments).
+{{< /admonition >}}
--- a/content/en/beginner-install.md
+++ b/content/en/beginner-install.md
@@ -62,10 +62,12 @@ with `pip` by executing the following in a terminal/shell:
 
     python -m pip install scipy
 
+{{< admonition warning >}}
 You may see this recommended in tutorials or classes, but it is not
 recommended to install packages globally on your system.
 The recommended way to install SciPy with `pip` is to use a virtual
 environment - see [Installing with `pip`](./install.md#installing-with-pip).
+{{< /admonition >}}
 
 {{< admonition note >}}
 For more information on why this is not a recommended installation method,
--- a/content/en/beginner-install.md
+++ b/content/en/beginner-install.md
@@ -20,7 +20,8 @@ how to use a package manager, check out
 
 To try out SciPy, you don't even need to install it!
 You can use SciPy in your browser at https://jupyter.org/try-jupyter/lab/ -
-just write `import scipy` in one of the notebook "cells" and hit play.
+just open a Python Notebook, then write `import scipy` in one of
+the notebook "cells" and hit play.
 
 For next steps, see [the SciPy user guide][scipy-user-guide].
 
@@ -46,8 +47,9 @@ Other options include:
   Windows, and Mac.
 
 {{< admonition note >}}
-Users in large, non-university institutions may want to read Anaconda's
-helpful blog on ["when is Anaconda free to use?"](https://www.anaconda.com/blog/update-on-anacondas-terms-of-service-for-academia-and-research)
+Anaconda is free to use for inviduals, universities, and companies smaller than
+200 employees. For more detail, see Anaconda's helpful blog on
+["when is Anaconda free to use?"](https://www.anaconda.com/blog/update-on-anacondas-terms-of-service-for-academia-and-research)
 {{< /admonition >}}
 
 After installing a scientific Python distribution,
@@ -58,6 +60,7 @@ see next steps in [the SciPy user guide][scipy-user-guide].
 ## Installing globally with `pip`
 
 If you already have Python installed, you can install SciPy globally
+(for all users on the system)
 with `pip` by executing the following in a terminal/shell:
 
     python -m pip install scipy
--- a/content/en/beginner-install.md
+++ b/content/en/beginner-install.md
@@ -6,7 +6,7 @@ sidebar: false
 {{< admonition tip >}}
 This is the beginner installation guide.
 If you are comfortable with using a terminal and happy to learn
-how to use a package manager, check out 
+how to use a package manager, check out
 [the main installation guide](./install.md)!
 {{< /admonition >}}
 
@@ -35,6 +35,7 @@ most commonly used scientific Python tools.
 [Anaconda](https://www.anaconda.com/download/) works on Windows, Mac,
 and Linux, and is best suited to beginning users.
 Other options include:
+
 - [WinPython](https://winpython.github.io): Another free distribution
   including scientific packages and the Spyder IDE; Windows only, but
   more actively maintained and supports the latest Python 3 versions.
--- a/content/en/beginner-install.md
+++ b/content/en/beginner-install.md
@@ -40,8 +40,7 @@ and Linux, and is best suited to beginning users.
 Other options include:
 
 - [WinPython](https://winpython.github.io): Another free distribution
-  including scientific packages and the Spyder IDE; Windows only, but
-  more actively maintained and supports the latest Python 3 versions.
+  including scientific packages and the Spyder IDE; Windows only.
 - [Pyzo](https://pyzo.org): A free distribution based on Anaconda
   and the IEP interactive development environment; Supports Linux,
   Windows, and Mac.
@@ -59,8 +58,7 @@ see next steps in [the SciPy user guide][scipy-user-guide].
 
 ## Installing globally with `pip`
 
-If you already have Python installed, you can install SciPy globally
-(for all users on the system)
+If you already have Python installed, you can install SciPy
 with `pip` by executing the following in a terminal/shell:
 
     python -m pip install scipy
--- a/content/en/beginner-install.md
+++ b/content/en/beginner-install.md
@@ -64,10 +64,9 @@ with `pip` by executing the following in a terminal/shell:
     python -m pip install scipy
 
 {{< admonition warning >}}
-You may see this recommended in tutorials or classes, but it is not
-recommended to install packages globally on your system.
-The recommended way to install SciPy with `pip` is to use a virtual
-environment - see [Installing with `pip`](./install.md#installing-with-pip).
+You may see this recommended in tutorials or classes, but the recommended
+way to install SciPy with `pip` is to use a virtual environment -
+see [Installing with `pip`](./install.md#installing-with-pip).
 {{< /admonition >}}
 
 {{< admonition note >}}
--- a/content/en/beginner-install.md
+++ b/content/en/beginner-install.md
@@ -24,6 +24,8 @@ just write `import scipy` in one of the notebook "cells" and hit play.
 
 For next steps, see [the SciPy user guide][scipy-user-guide].
 
+[scipy-user-guide]: https://docs.scipy.org/doc/scipy/tutorial/
+
 <a name="distributions"></a>
 
 ## Scientific Python Distributions
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -46,7 +46,7 @@ a Python package manager.
         cd try-scipy
 
     {{< admonition hint >}}
-    The second command changes directory into the directory of your project.
+The second command changes directory into the directory of your project.
     {{< /admonition >}}
 
 3.  Install Python:
@@ -54,7 +54,7 @@ a Python package manager.
         uv python install
 
     {{< admonition note >}}
-    If you have already installed Python, this step is optional.
+If you have already installed Python, this step is optional.
     {{< /admonition >}}
 
 4.  Add SciPy to your project:
@@ -101,7 +101,7 @@ tool [`pixi`] are very similar to the steps for `uv`:
         uv add scipy
 
     {{< admonition note >}}
-    This step also adds Python from conda-forge.
+This step also adds Python from conda-forge.
     {{< /admonition >}}
 
 4.  Try out SciPy!
@@ -126,7 +126,7 @@ but lack some reproducibility benefits of project-based workflows.
 2. Create and activate a virtual environment with `venv`.
 
    {{< admonition hint >}}
-   See [the tutorial in the Python Packaging User Guide][venv-guide].
+See [the tutorial in the Python Packaging User Guide][venv-guide].
    {{< /admonition >}}
 
 [venv-guide]: https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -126,10 +126,8 @@ but lack some reproducibility benefits of project-based workflows.
 2. Create and activate a virtual environment with `venv`.
 
    {{< admonition hint >}}
-See [the tutorial in the Python Packaging User Guide][venv-guide].
+See [the tutorial in the Python Packaging User Guide](https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments).
    {{< /admonition >}}
-
-[venv-guide]: https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments
 
 3.  Install SciPy, using [`pip`]:
 
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -3,125 +3,89 @@ title: Installation
 sidebar: false
 ---
 
-Installation methods include:
+The recommended method of installing SciPy depends on your preferred workflow.
+The common workflows can roughly be broken down into the following
+categories:
 
-- [Python Distributions](#distributions)
-- [PyPI (`pip`, `uv`)](#pypi-install)
-- [conda-forge (`conda`, `mamba`, `pixi`)](#conda-forge-install)
-- [System Package Managers](#system-package-manager)
-- [Source](#source)
-
-Methods differ in ease of use, coverage, maintenance of old versions,
-system-wide versus local environment use, and control. When installing
-from PyPI or with Conda, you can control the package versions for a specific
-project to prevent conflicts. Conda also controls non-Python packages,
-like MKL or HDF5. System package managers, like `apt-get`, install
-across the entire computer, often have older versions, and don't have
-as many available versions. Source compilation is much more difficult
-but is necessary for debugging and development. If you don't know which
-installation method you need or prefer, we recommend the Scientific
-Python Distribution [Anaconda](https://www.anaconda.com/download/).
-
-Users in large, non-university institutions may want to read Anaconda's
-helpful blog on ["when is Anaconda free to use?"][anaconda-blog]
-
-[anaconda-blog]: https://www.anaconda.com/blog/update-on-anacondas-terms-of-service-for-academia-and-research
+- [Environment-based (e.g. `pip`, `conda`)](#environment-based) (the traditional workflow)
+- [Project-based (e.g. `uv`, `pixi`)](#project-based) (recommended for new users)
+- [System package managers](#system-package-managers) (simplest)
+- [Building from source](#building-from-source) (for debugging and development)
 
 To install SciPy with [static type stubs],
 see [Installing with type stubs](#type-stubs).
 
 [static type stubs]: https://typing.readthedocs.io/en/latest/guides/libraries.html
 
-<a name="distributions"></a>
+<a name="environment-based"></a>
 
-## Scientific Python Distributions (recommended for new users)
+## Environment-based workflows
 
-Python distributions provide the language itself, along with the most
-commonly used packages and tools. These downloadable files require
-little configuration, work on almost all setups, and provide all the
-commonly used scientific Python tools.
+In environment-based workflows, you install packages into an environment, which you
+can activate and deactivate.
 
-[Anaconda](https://www.anaconda.com/download/) works on Windows, Mac,
-and Linux, provides over 1,500 Python packages, and is used by over 15
-million people. Anaconda is best suited to beginning users; it provides
-a large collection of libraries all in one.
-
-Other options include:
-
-- [WinPython](https://winpython.github.io): Another free distribution
-  including scientific packages and the Spyder IDE; Windows only, but
-  more actively maintained and supports the latest Python 3 versions.
-- [Pyzo](https://pyzo.org): A free distribution based on Anaconda
-  and the IEP interactive development environment; Supports Linux,
-  Windows, and Mac.
-
-If you already have Python installed, you can
-[install SciPy with `pip`](#pip-install).
-For more advanced users who will need to install or upgrade regularly,
-you may prefer to use a dedicated package manager like [`uv`]
-(see [Installing with `uv`](#uv-install)),
-[Conda], or [`pixi`] (see [Installing from conda-forge](#conda-forge-install)).
-
-[`uv`]: https://docs.astral.sh/uv/
-[Conda]: https://docs.conda.io/projects/conda/en/latest/index.html
-[`pixi`]: https://pixi.sh/latest/
-
-<a name="pypi-install"></a>
-
-## Installing from PyPI
-
-SciPy is [available on PyPI](https://pypi.org/project/scipy/), the Python Package Index.
-
-<a name="pip-install"></a>
-
-### Installing with `pip`
-
-You can install SciPy from PyPI with `pip`:
+If you have installed Python, you can create a virtual environment with a tool such
+as `venv` - see [the tutorial in the Python Packaging User Guide][venv-guide].
+You can then install [SciPy from PyPI], the Python Package Index, with [`pip`]:
 
     python -m pip install scipy
 
-<a name="uv-install"></a>
+[venv-guide]: https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments
+[SciPy from PyPI]: https://pypi.org/project/scipy/
+[`pip`]: https://pip.pypa.io/en/stable/getting-started/
 
-### Installing with `uv`
+Alternatively, you can install [SciPy from conda-forge] using a [Conda]-based
+environment manager. You can install Python itself, as well as packages in
+languages other than Python, as Conda packages.
 
-You can also add SciPy from PyPI to a `uv` project:
+[Miniforge] is the recommended way to install `conda` and [`mamba`],
+two Conda-based environment managers. You can then install SciPy as follows:
+
+    conda install scipy # or
+    mamba install scipy
+
+[SciPy from conda-forge]: https://anaconda.org/conda-forge/scipy
+[Miniforge]: https://conda-forge.org/download/
+[Conda]: https://docs.conda.io/projects/conda/en/latest/index.html
+[`mamba`]: https://mamba.readthedocs.io/en/latest/
+
+<a name="project-based"></a>
+
+## Project-based workflows
+
+In project-based workflows, a project is a directory containing a manifest
+file describing the project, a lock-file describing the exact dependencies
+of the project, and the project's (potentially multiple) environments.
+These workflows provide reproducibility benefits over environment-based workflows.
+
+You can add [SciPy from PyPI] to a [`uv`] project:
 
     uv add scipy
 
-<a name="conda-forge-install"></a>
+You can also install Python itself with `uv`.
 
-## Installing from conda-forge
+[`uv`]: https://docs.astral.sh/uv/
 
-### Installing with `conda` or `mamba`
+If you work with non-Python packages, you may prefer to install SciPy as
+a Conda package, so that you can use the same workflow for packages which
+are not available on PyPI.
 
-[`Miniforge`] is the recommended way to install the `conda`
-(and `mamba`, a faster `conda` alternative) package manager.
-You can then install SciPy as follows:
-
-    mamba install scipy
-
-[`Miniforge`]: https://conda-forge.org/download/
-
-### Installing with `pixi`
-
-You can also add SciPy from conda-forge to a [`pixi`] project:
+You can add [SciPy from conda-forge] to a [`pixi`] project:
 
     pixi add scipy
 
-## Installing from the Anaconda Repository
+Pixi can also install packages from PyPI, using `uv` under the hood.
 
-If you instead install `conda` via [`Miniconda`], you can install
-[SciPy from the Anaconda repository](https://anaconda.org/anaconda/scipy):
+[`pixi`]: https://pixi.sh/latest/
 
-    conda install scipy
+<a name="system-package-managers"></a>
 
-<a name="system-package-manager"></a>
-
-## Install system-wide via a system package manager
+## Installing system-wide via a system package manager
 
 System package managers can install the most common Python packages.
 They install packages for the entire computer, often use older versions,
-and don't have as many available versions.
+and don't have as many available versions. They are not the recommended
+installation method, but they are the simplest.
 
 ### Ubuntu and Debian
 
@@ -142,14 +106,16 @@ macOS doesn't have a preinstalled package manager, but you can install
 
     brew install scipy
 
-<a name="source"></a>
+<a name="building-from-source"></a>
 
-## Source packages
+## Building from source
 
 A word of warning: building SciPy from source can be a nontrivial exercise. We
 recommend using binaries instead if those are available for your platform.
 For details on how to build from source, see
-[this guide in the SciPy docs](https://scipy.github.io/devdocs/building/index.html).
+[the building from source guide in the SciPy docs][building-docs].
+
+[building-docs]: https://scipy.github.io/devdocs/building/index.html
 
 <a name="type-stubs"></a>
 
@@ -161,7 +127,7 @@ install `scipy-stubs` in the same way:
 
     python -m pip install scipy-stubs # or
     uv add scipy-stubs # or
-    mamba install scipy-stubs # or
+    conda install scipy-stubs # or
     pixi add scipy-stubs
 
 <!---
@@ -174,8 +140,11 @@ package on conda-forge.
 To get a specific version `x.y.z` of SciPy (such as `1.14.1`),
 you should install version `x.y.z.*` below:
 
-    python -m pip install 'scipy-stubs[scipy]' # or
+    python -m pip install "scipy-stubs[scipy]" # or
     # versions for illustrative purposes
-    uv add 'scipy-stubs[scipy]==1.14.*' # or
-    mamba install 'scipy-typed==1.14.1.*' # or
-    pixi add 'scipy-typed==1.15.0.*'
+    uv add "scipy-stubs[scipy]==1.14.1.*" # or
+    conda install "scipy-typed>=1.14" # or
+    pixi add "scipy-typed=1.15.0.*"
+
+Please direct questions about static typing support to
+[the `scipy-stubs` repo](https://github.com/jorenham/scipy-stubs).
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -70,9 +70,7 @@ you may prefer to use a dedicated package manager like [`uv`]
 
 # Installing from PyPI
 
-SciPy is [available on PyPI](PyPI-scipy), the Python Package Index.
-
-[PyPI-scipy]: https://pypi.org/project/scipy/
+SciPy is [available on PyPI](https://pypi.org/project/scipy/), the Python Package Index.
 
 <a name="pip-install"></a>
 
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -35,6 +35,8 @@ a Python package manager.
 
 [`uv`]: https://docs.astral.sh/uv/
 
+<!-- prettier-ignore-start -->
+
 1. Install `uv`, following [the instructions in the `uv` documentation][install-uv].
 
 [install-uv]: https://docs.astral.sh/uv/getting-started/installation/
@@ -56,18 +58,20 @@ The second command changes directory into the directory of your project.
 This will automatically install Python if you don't already have it installed!
     {{< /admonition >}}
 
-{{< admonition tip >}}
+    {{< admonition tip >}}
 You can install other Python libraries in the same way, e.g.
 
     uv add matplotlib
 
-{{< /admonition >}}
+    {{< /admonition >}}
 
 4.  Try out SciPy!
 
         uv run python
 
     This will launch a Python interpreter session, from which you can `import scipy`.
+
+<!-- prettier-ignore-end -->
 
 See next steps in [the SciPy user guide][scipy-user-guide].
 
@@ -137,6 +141,8 @@ but lack some reproducibility benefits of project-based workflows.
 
 ### Installing with `pip`
 
+<!-- prettier-ignore-start -->
+
 1.  [Install Python](https://www.python.org/downloads/).
 
 2.  Create and activate a virtual environment with `venv`.
@@ -148,6 +154,8 @@ See [the tutorial in the Python Packaging User Guide](https://packaging.python.o
 3.  Install SciPy, using [`pip`]:
 
         python -m pip install scipy
+
+<!-- prettier-ignore-end -->
 
 [`pip`]: https://pip.pypa.io/en/stable/getting-started/
 
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -53,7 +53,7 @@ The second command changes directory into the directory of your project.
 This will automatically install Python if you don't already have it installed!
     {{< /admonition >}}
 
-{{< admonition note >}}
+{{< admonition tip >}}
 You can install other Python libraries in the same way, e.g.
 
     uv add matplotlib
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -40,7 +40,7 @@ a Python package manager.
 
 [install-uv]: https://docs.astral.sh/uv/getting-started/installation/
 
-2. Create a new project in a new subdirectory, by executing the following in a terminal:
+2.  Create a new project in a new subdirectory, by executing the following in a terminal:
 
         uv init try-scipy
         cd try-scipy
@@ -49,7 +49,7 @@ a Python package manager.
     The second command changes directory into the directory of your project.
     {{< /admonition >}}
 
-3. Install Python:
+3.  Install Python:
 
         uv python install
 
@@ -57,11 +57,11 @@ a Python package manager.
     If you have already installed Python, this step is optional.
     {{< /admonition >}}
 
-4. Add SciPy to your project:
+4.  Add SciPy to your project:
 
         uv add scipy
 
-5. Try out SciPy!
+5.  Try out SciPy!
 
         uv run python
 
@@ -91,12 +91,12 @@ tool [`pixi`] are very similar to the steps for `uv`:
 
 [install-pixi]: https://pixi.sh/latest/
 
-2. Create a new project in a new subdirectory:
+2.  Create a new project in a new subdirectory:
 
         pixi init try-scipy
         cd try-scipy
 
-4. Add SciPy to your project:
+3.  Add SciPy to your project:
 
         uv add scipy
 
@@ -104,7 +104,7 @@ tool [`pixi`] are very similar to the steps for `uv`:
     This step also adds Python from conda-forge.
     {{< /admonition >}}
 
-5. Try out SciPy!
+4.  Try out SciPy!
 
         pixi run python
 
@@ -125,13 +125,13 @@ but lack some reproducibility benefits of project-based workflows.
 
 2. Create and activate a virtual environment with `venv`.
 
-    {{< admonition hint >}}
-    See [the tutorial in the Python Packaging User Guide][venv-guide].
-    {{< /admonition >}}
+   {{< admonition hint >}}
+   See [the tutorial in the Python Packaging User Guide][venv-guide].
+   {{< /admonition >}}
 
 [venv-guide]: https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments
 
-3. Install SciPy, using [`pip`]:
+3.  Install SciPy, using [`pip`]:
 
         python -m pip install scipy
 
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -3,6 +3,13 @@ title: Installation
 sidebar: false
 ---
 
+{{< admonition tip >}}
+This page assumes that you are comfortable with using a terminal and happy to learn
+how to use a package manager. If you are a beginner and just want to get started
+with SciPy as quickly as possible, check out
+[the beginner installation guide](./beginner-install.md)!
+{{< /admonition >}}
+
 The recommended method of installing SciPy depends on your preferred workflow.
 The common workflows can roughly be broken down into the following
 categories:
@@ -23,11 +30,7 @@ see [Installing with type stubs](#type-stubs).
 
 ### Installing with `uv`
 
-{{< admonition tip >}}
-If it is your first time installing a Python library, start here!
-{{< /admonition >}}
-
-Here is a step-by-step guide to setting up a project to try SciPy, with [`uv`],
+Here is a step-by-step guide to setting up a project to use SciPy, with [`uv`],
 a Python package manager.
 
 [`uv`]: https://docs.astral.sh/uv/
@@ -69,6 +72,19 @@ You can install other Python libraries in the same way, e.g.
 See next steps in [the SciPy user guide][scipy-user-guide].
 
 [scipy-user-guide]: https://docs.scipy.org/doc/scipy/tutorial/
+
+{{< admonition note >}}
+
+After rebooting your computer, you'll want to navigate to your `try-scipy`
+project directory and execute `uv run python` to drop back into a Python interpreter
+with SciPy importable.
+To execute a Python script, you can use `uv run myscript.py`.
+
+Read more at [the uv guide to working on projects][uv-projects].
+
+[uv-projects]: https://docs.astral.sh/uv/guides/projects/
+
+{{< /admonition >}}
 
 ### Installing with `pixi`
 
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -147,4 +147,4 @@ you should install version `x.y.z.*` below:
     pixi add "scipy-typed=1.15.0.*"
 
 Please direct questions about static typing support to
-[the `scipy-stubs` repo](https://github.com/jorenham/scipy-stubs).
+[the `scipy-stubs` GitHub repository](https://github.com/jorenham/scipy-stubs).
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -98,7 +98,7 @@ tool [`pixi`] are very similar to the steps for `uv`:
 
 3.  Add SciPy to your project:
 
-        uv add scipy
+        pixi add scipy
 
     {{< admonition note >}}
 This step also adds Python from conda-forge.
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -179,4 +179,3 @@ you should install version `x.y.z.*` below:
     uv add 'scipy-stubs[scipy]==1.14.*' # or
     mamba install 'scipy-typed==1.14.1.*' # or
     pixi add 'scipy-typed==1.15.0.*'
-
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -111,7 +111,8 @@ macOS doesn't have a preinstalled package manager, but you can install
 ## Building from source
 
 A word of warning: building SciPy from source can be a nontrivial exercise. We
-recommend using binaries instead if those are available for your platform.
+recommend using binaries instead if those are available for your platform
+via one of the above methods.
 For details on how to build from source, see
 [the building from source guide in the SciPy docs][building-docs].
 
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -34,7 +34,7 @@ see [Installing with type stubs](#type-stubs).
 
 <a name="distributions"></a>
 
-# Scientific Python Distributions (recommended for new users)
+## Scientific Python Distributions (recommended for new users)
 
 Python distributions provide the language itself, along with the most
 commonly used packages and tools. These downloadable files require
@@ -68,13 +68,13 @@ you may prefer to use a dedicated package manager like [`uv`]
 
 <a name="pypi-install"></a>
 
-# Installing from PyPI
+## Installing from PyPI
 
 SciPy is [available on PyPI](https://pypi.org/project/scipy/), the Python Package Index.
 
 <a name="pip-install"></a>
 
-## Installing with `pip`
+### Installing with `pip`
 
 You can install SciPy from PyPI with `pip`:
 
@@ -82,7 +82,7 @@ You can install SciPy from PyPI with `pip`:
 
 <a name="uv-install"></a>
 
-## Installing with `uv`
+### Installing with `uv`
 
 You can also add SciPy from PyPI to a `uv` project:
 
@@ -90,9 +90,9 @@ You can also add SciPy from PyPI to a `uv` project:
 
 <a name="conda-forge-install"></a>
 
-# Installing from conda-forge
+## Installing from conda-forge
 
-## Installing with `conda` or `mamba`
+### Installing with `conda` or `mamba`
 
 [`Miniforge`] is the recommended way to install the `conda`
 (and `mamba`, a faster `conda` alternative) package manager.
@@ -102,13 +102,13 @@ You can then install SciPy as follows:
 
 [`Miniforge`]: https://conda-forge.org/download/
 
-## Installing with `pixi`
+### Installing with `pixi`
 
 You can also add SciPy from conda-forge to a [`pixi`] project:
 
     pixi add scipy
 
-# Installing from the Anaconda Repository
+## Installing from the Anaconda Repository
 
 If you instead install `conda` via [`Miniconda`], you can install
 [SciPy from the Anaconda repository](https://anaconda.org/anaconda/scipy):
@@ -117,25 +117,25 @@ If you instead install `conda` via [`Miniconda`], you can install
 
 <a name="system-package-manager"></a>
 
-# Install system-wide via a system package manager
+## Install system-wide via a system package manager
 
 System package managers can install the most common Python packages.
 They install packages for the entire computer, often use older versions,
 and don't have as many available versions.
 
-## Ubuntu and Debian
+### Ubuntu and Debian
 
 Using `apt-get`:
 
     sudo apt-get install python3-scipy
 
-## Fedora
+### Fedora
 
 Using `dnf`:
 
     sudo dnf install python3-scipy
 
-## macOS
+### macOS
 
 macOS doesn't have a preinstalled package manager, but you can install
 [Homebrew](https://brew.sh/) and use it to install SciPy (and Python itself):
@@ -144,7 +144,7 @@ macOS doesn't have a preinstalled package manager, but you can install
 
 <a name="source"></a>
 
-# Source packages
+## Source packages
 
 A word of warning: building SciPy from source can be a nontrivial exercise. We
 recommend using binaries instead if those are available for your platform.
@@ -153,7 +153,7 @@ For details on how to build from source, see
 
 <a name="type-stubs"></a>
 
-# Installing with type stubs
+## Installing with type stubs
 
 Static type stubs are available via a separate package, `scipy-stubs`, on
 PyPI and conda-forge. If you have already installed SciPy, you just need to
--- a/content/en/install.md
+++ b/content/en/install.md
@@ -7,47 +7,15 @@ The recommended method of installing SciPy depends on your preferred workflow.
 The common workflows can roughly be broken down into the following
 categories:
 
-- [Environment-based (e.g. `pip`, `conda`)](#environment-based) (the traditional workflow)
 - [Project-based (e.g. `uv`, `pixi`)](#project-based) (recommended for new users)
-- [System package managers](#system-package-managers) (simplest)
+- [Environment-based (e.g. `pip`, `conda`)](#environment-based) (the traditional workflow)
+- [System package managers](#system-package-managers) (not recommended)
 - [Building from source](#building-from-source) (for debugging and development)
 
 To install SciPy with [static type stubs],
 see [Installing with type stubs](#type-stubs).
 
 [static type stubs]: https://typing.readthedocs.io/en/latest/guides/libraries.html
-
-<a name="environment-based"></a>
-
-## Environment-based workflows
-
-In environment-based workflows, you install packages into an environment, which you
-can activate and deactivate.
-
-If you have installed Python, you can create a virtual environment with a tool such
-as `venv` - see [the tutorial in the Python Packaging User Guide][venv-guide].
-You can then install [SciPy from PyPI], the Python Package Index, with [`pip`]:
-
-    python -m pip install scipy
-
-[venv-guide]: https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments
-[SciPy from PyPI]: https://pypi.org/project/scipy/
-[`pip`]: https://pip.pypa.io/en/stable/getting-started/
-
-Alternatively, you can install [SciPy from conda-forge] using a [Conda]-based
-environment manager. You can install Python itself, as well as packages in
-languages other than Python, as Conda packages.
-
-[Miniforge] is the recommended way to install `conda` and [`mamba`],
-two Conda-based environment managers. You can then install SciPy as follows:
-
-    conda install scipy # or
-    mamba install scipy
-
-[SciPy from conda-forge]: https://anaconda.org/conda-forge/scipy
-[Miniforge]: https://conda-forge.org/download/
-[Conda]: https://docs.conda.io/projects/conda/en/latest/index.html
-[`mamba`]: https://mamba.readthedocs.io/en/latest/
 
 <a name="project-based"></a>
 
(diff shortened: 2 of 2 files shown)


content/en/install.md
```diff
@@ -50,43 +55,87 @@ Other options include:
...
+You can also add SciPy from conda-forge to a [`pixi`] project:
+
+    pixi add scipy
+
+## Installing from the Anaconda Repository
+
+If you instead install `conda` via [`Miniconda`], you can install
+[SciPy from the Anaconda repository](https://anaconda.org/anaconda/scipy):
 
     conda install scipy
```
lucascolley, line 117: do we need to keep this?

content/en/install.md
```diff
@@ -50,43 +55,87 @@ Other options include:
   and the IEP interactive development environment; Supports Linux,
   Windows, and Mac.
 
+If you already have Python installed, you can
+[install SciPy with `pip`](#pip-install).
```
lucascolley, line 59: not necessarily true but I think it's close enough to true for here

content/en/install.md
```diff
@@ -50,43 +55,87 @@ Other options include:
...
+<a name="pypi-install"></a>
+
+## Installing from PyPI
+
+SciPy is [available on PyPI](https://pypi.org/project/scipy/), the Python Package Index.
+
 <a name="pip-install"></a>
 
-# Installing with Pip
+### Installing with `pip`
```
lucascolley, line 77: perhaps better without these subheadings?

jorenham:
Good stuff! Maybe it could be a good idea to mention the `scipy-stubs` repo, so that stuff like bug reports ends up at the right place?

content/en/install.md
```diff
@@ -3,28 +3,38 @@ title: Installation
...
 like MKL or HDF5. System package managers, like `apt-get`, install
-across the entire computer, often have older versions, and don\'t have
+across the entire computer, often have older versions, and don't have
 as many available versions. Source compilation is much more difficult
-but is necessary for debugging and development. If you don\'t know which
+but is necessary for debugging and development. If you don't know which
 installation method you need or prefer, we recommend the Scientific
 Python Distribution [Anaconda](https://www.anaconda.com/download/).
 
+Users in large, non-university institutions may want to read Anaconda's
```
tupui, line 25: I won't block, but I disagree with putting this here. Users to which this really applies need to do such analysis anyway for all the tools they use, have procurement teams, etc.

This only sends a bad message for Anaconda which has been giving a lot (and still do) to this community.

content/en/install.md
```diff
@@ -50,43 +55,87 @@ Other options include:
   and the IEP interactive development environment; Supports Linux,
   Windows, and Mac.
 
+If you already have Python installed, you can
+[install SciPy with `pip`](#pip-install).
+For more advanced users who will need to install or upgrade regularly,
+you may prefer to use a dedicated package manager like [`uv`]
+(see [Installing with `uv`](#uv-install)),
+[Conda], or [`pixi`] (see [Installing from conda-forge](#conda-forge-install)).
```
tupui, line 63: -1 for Pixi. Looking at how things are moving right now, uv is going to completely shadow such alternative solution.

content/en/install.md
```diff
@@ -50,43 +55,87 @@ Other options include:
...
+<a name="conda-forge-install"></a>
 
-# Installing via Conda
+## Installing from conda-forge
 
-You can install SciPy from the `defaults` or `conda-forge` channels with `conda`:
+### Installing with `conda` or `mamba`
+
+[`Miniforge`] is the recommended way to install the `conda`
+(and `mamba`, a faster `conda` alternative) package manager.
```
tupui, line 98: This is becoming less true as the faster engine got into conda. Maybe worth asking someone from the conda team.

lucascolley: Good point, @jaimergp perhaps you could clarify whether it seems worthwhile to distinguish between `conda` and `mamba` here?

content/en/install.md
lucascolley: I think it is worthwhile to mention it, at least while uv cannot handle conda packages (https://github.com/astral-sh/uv/issues/1703). In my experience, it is nice to have one tool for everything, rather than having to use something else for non-PyPI packages.

content/en/install.md
lucascolley: I think it is a useful pointer for us to include, and I don't think that linking out to Anaconda's own blog post sends a bad message (if anything, it could help make users who should be paying Anaconda aware that they need to!)

I think there were valid concerns raised in  https://docs.google.com/document/d/1zfncPEtNgI86pm8AhJ50DCGSTexcEIuA6jUofy2yqvU/edit?tab=t.0#heading=h.exik8qo8laee about a situation which was pretty confusing (at least for me) to understand, and Anaconda's blog post helpfully explained it.

But I also won't block, if folks would rather not include this :)

content/en/install.md
tupui: I left an upvote on the issue and pinged Charlie to get some visibility. He helped us gladly in the past with Ruff 🤞

content/en/install.md
tupui: I remember being against that doc and NF having an opinion 😉 This is political at best, again users to which this applies know what to do. There won't be many large orgs which don't have a proper process around compliant tooling. People who don't pay do it on purpose and this note won't help.

Also from a practical point of view, such link tend to break as time pass. And the TOC can always change.

content/en/install.md
tupui: @zanieb 😃 since you asked. Thanks 🙏

zanieb: We're not actively working on support for installing things from Conda in uv, so I think it's reasonable to have an alternative suggestion. Whether you prefer to just suggest using `mamba` or `pixi` seems dependent on how important the "project" use-case is. My 2-cents is that project-focused users usually use PyPI while the Conda users are usually more "environment" focused — in which case a project-based approach offered by uv and pixi is less important.

content/en/install.md
```diff
@@ -50,43 +55,87 @@ Other options include:
...
+(see [Installing with `uv`](#uv-install)),
+[Conda], or [`pixi`] (see [Installing from conda-forge](#conda-forge-install)).
+
+[`uv`]: https://docs.astral.sh/uv/
+[Conda]: https://docs.conda.io/projects/conda/en/latest/index.html
+[`pixi`]: https://pixi.sh/latest/
+
+<a name="pypi-install"></a>
+
+## Installing from PyPI
```
zanieb, line 71: Is the _source_ of the installation the major thing to split on here? Or should it be more like:

- Installing into an environment
   - Using pip, uv, or conda
- Adding to your project
    - Using uv or pixi
- Installing system-wide
    - ...

content/en/install.md
zanieb: That's great to hear! I definitely think Pixi is great and that project-based workflows are better than maintaining imperative global environments :) I'm just not sure how common it is.

content/en/install.md
lucascolley: That sounds pretty good to me! Although it is more of a departure from the existing text

content/en/install.md
rgommers: I think we should just change the recommendations to only use Conda-forge. There is no reason to use `defaults` at this point if you're an individual user trying to figure this stuff out. So we get rid of https://www.anaconda.com/download/ and then we don't need to worry about the ToS drama.

lucascolley: > So we get rid of https://www.anaconda.com/download/

Are you suggesting removing the whole "Scientific Python Distributions" section?

content/en/install.md
rgommers: I don't think it's constructive at all to say no to `pixi` because of `uv. Basically, `uv` is a very nice upgrade from `pip` & co, while `pixi is a very nice upgrade from `conda` & co. 

> I guess I am bucking the trend by being a conda & project-workflow enjoyer then :)

I don't think you are. Many maintainers I talk to are excited about Pixi, because it gives more reproducible workflows. And as soon as you include compiled code, it can give you reproducibility in a way that PyPI-only tooling simply cannot.

content/en/install.md
rgommers: Yes

content/en/install.md
lucascolley: If I understand correctly, (broadly speaking, from a non-expert user perspective)

- `conda` now provides the speed and effectively all of the benefits `mamba` used to have
- mamba 2.0 makes `mamba` very similar to `micromamba`, which is still more lightweight in a sense than `conda`?

lucascolley:
> Maybe it could be a good idea to mention the scipy-stubs repo, so that stuff like bug reports ends up at the right place?

I think that would be okay, although maybe sufficient to add a notice to the SciPy bug report template.

content/en/install.md
```diff
@@ -3,101 +3,148 @@ title: Installation
...
+environment manager. You can install Python itself, as well as packages in
+languages other than Python, as Conda packages.
 
-# Installing with Pip
+[Miniforge] is the recommended way to install `conda` and [`mamba`],
+two Conda-based environment managers. You can then install SciPy as follows:
 
-You can install SciPy from PyPI with `pip`:
+    conda install scipy # or
+    mamba install scipy
```
lucascolley, line 45: I've left `mamba` in for now, but perhaps we don't need to mention it? It is discoverable from the Miniforge homepage, and to be honest I'm still not clear on how mamba 2.0 compares to conda.

lucascolley (commented):
Thanks all for the comments! I have addressed all feedback as follows:

- Removed section on Python Distributions
- Discerned primarily between environment-based and project-based workflows, rather than PyPI vs conda-forge. Marked environment-based as "the traditional workflow", project-based as "recommended for new users", and "system package managers" as "simplest" (but not recommended).
- Removed existing discussion of pip vs conda vs system vs source
- Linked to the external docs for all tools

Let me know what you think!

content/en/install.md
```diff
@@ -3,101 +3,148 @@ title: Installation
...
     brew install scipy
 
-<a name="source"></a>
+<a name="building-from-source"></a>
 
-# Source packages
+## Building from source
 
 A word of warning: building SciPy from source can be a nontrivial exercise. We
 recommend using binaries instead if those are available for your platform.
```
lucascolley, line 114:
```suggestion
recommend using binaries instead if those are available for your platform
via one of the above methods.
```

jorenham (approved):

content/en/install.md
```diff
@@ -3,101 +3,148 @@ title: Installation
...
 
     sudo dnf install python3-scipy
 
-## macOS
+### macOS
 
 macOS doesn't have a preinstalled package manager, but you can install
 [Homebrew](https://brew.sh/) and use it to install SciPy (and Python itself):
 
     brew install scipy
```
hameerabbasi, line 107: `brew` is falling more and more out of favour as it has a horrible permissions model: Only one user can use it on one system OOTB. I'd recommend MacPorts if we want something Mac-specific.

tupui: I did not hear MacPort being mentioned in years. Both around OSS and professionally. Do you have any supporting material to back that statement up? 😅

content/en/install.md
jaimergp: I'd mention conda and micromamba. Ultimately it's up to the user though, and their muscle memory. Mamba 2.0 is like micromamba but dynamically linked.

content/en/install.md
hameerabbasi: About the multi-user issue with Homebrew:

> If you need to run Homebrew in a multi-user environment, consider creating a separate user account specifically for use of Homebrew.

From [the docs](https://docs.brew.sh/FAQ#why-does-homebrew-say-sudo-is-bad).

A lot of workarounds for per-user installs are unsupported: [1](https://docs.brew.sh/Installation#untar-anywhere-unsupported) [2](https://docs.brew.sh/Installation#multiple-installations-unsupported).

I don't use MacPorts personally; mostly `pixi` these days; but Homebrew is a potential footgun for the uninitiated.

content/en/install.md
rgommers: I don't agree, `mamba` 2.0 still has the same UX as `mamba` 1.x - so like `conda` not like `minimamba`, see https://mamba.readthedocs.io/en/latest/user_guide/mamba.html#mamba-vs-conda-clis), it's just further improved. And it has features `conda` doesn't have (e.g. `repoquery` is super useful). CLI output is also better, and I think (but am not sure) that downloads are faster as well.

So I'd much prefer to keep `mamba`, as the reasoning for adding it hasn't really changed - it's still a better `conda`.

rgommers: You really have to separate thinking about this from a user perspective vs. from an internals perspective. Dynamic linking is irrelevant to a user, unless it somehow translates into observable differences in behavior.

content/en/install.md
rgommers: MacPorts is still a thing, but we should keep Homebrew. It's the most popular package manager on macOS, and the limitation of no per-user installs on a shared machine is pretty niche, I haven't seen that come up as an issue before. We can add MacPorts of course. Both aren't recommended either way I'd say, since they're similar to system package managers. Users are better off knowing how to create environments (whether venvs or conda envs).

hameerabbasi: > the limitation of no per-user installs on a shared machine is pretty niche

That's one limitation -- another is that the global installation cannot be ergonomically used by multiple users. The only supported way to do that is to imitate another user; as the permissions are tied to a user, not a group.

content/en/install.md
lucascolley: In which case, I'm inclined to leave this text as is. Given https://github.com/mamba-org/mamba/pull/3610, it seems that micromamba is intended to be more developer facing than user facing now. cc @jjerphan for visibility.

jjerphan: In a nutshell:
 - `mamba` and `micromamba` are the same program, only their linkage change (`mamba` is dynamically linked, whilst `micromamba` is nearly totally-statically linked)
 - `conda` now uses `libmamba` to solve environments (via `conda-libmamba-solver`), so performance is on par with `mamba`
 - `mamba` and `conda` have differences in features
 - `mamba` is targeting conda-forge by default now

I would recommend using `mamba`.

See [the differences between `mamba` and `micromamba`](https://github.com/mamba-org/mamba/?tab=readme-ov-file#mamba-vs-micromamba).

lucascolley:
shall we give a week or so for any more comments here? This is ready from my side.

content/en/install.md
```diff
@@ -3,101 +3,149 @@ title: Installation
...
-million people. Anaconda is best suited to beginning users; it provides
-a large collection of libraries all in one.
+In environment-based workflows, you install packages into an environment, which you
+can activate and deactivate.
 
-For more advanced users who will need to install or upgrade regularly,
-[Mambaforge](https://github.com/conda-forge/miniforge#mambaforge) is a more
-suitable way to install the `conda` (and `mamba`, a faster `conda` alternative)
-package manager.
+If you have installed Python, you can create a virtual environment with a tool such
```
stefanv, line 27: "a tool such as" --- this suggests an external tool. I'd make explicit, e.g. "Python's built-in venv manager" or "Python's built-in venv tool".

content/en/install.md
```diff
@@ -3,101 +3,149 @@ title: Installation
...
+The common workflows can roughly be broken down into the following
+categories:
 
-- [Distributions](#distributions)
-- [pip](#pip-install)
-- [conda](#pip-install)
-- [Package Manager](#package_manager)
-- [Source](#source)
+- [Environment-based (e.g. `pip`, `conda`)](#environment-based) (the traditional workflow)
+- [Project-based (e.g. `uv`, `pixi`)](#project-based) (recommended for new users)
```
stefanv, line 11: The installations page should be for beginners, and they likely won't follow the project-based route. These are people who just want to get "import scipy" to work.

I think an explicit recommendation is a good idea, because newcomers should read this and immediately be able to say: "ah! this is how I install SciPy" without too many questions. I.e., we need to reduce the number of choices they have to make to as close to zero as possible.

If you want to be more nuanced about it, you could say, for env based: "recommended for beginners", and for project-based: "recommended for advanced users".

content/en/install.md
```diff
@@ -3,101 +3,149 @@ title: Installation
...
-  and the IEP interactive development environment; Supports Linux,
-  Windows, and Mac.
+[venv-guide]: https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments
+[SciPy from PyPI]: https://pypi.org/project/scipy/
+[`pip`]: https://pip.pypa.io/en/stable/getting-started/
 
-<a name="pip-install"></a>
+Alternatively, you can install [SciPy from conda-forge] using a [Conda]-based
+environment manager. You can install Python itself, as well as packages in
+languages other than Python, as Conda packages.
```
stefanv, line 39:
```suggestion
Alternatively, you can install [SciPy from conda-forge] using a [Conda]-based
environment manager. Conda is a full-fledged package manager, so you can also use it to install Python itself, compilers, and other languages.
```

Or, if you prefer different wording, the idea here is to make explicit the feature of conda being advertised.

content/en/install.md
```diff
@@ -3,101 +3,149 @@ title: Installation
...
+[SciPy from conda-forge]: https://anaconda.org/conda-forge/scipy
+[Miniforge]: https://conda-forge.org/download/
+[Conda]: https://docs.conda.io/projects/conda/en/latest/index.html
+[`mamba`]: https://mamba.readthedocs.io/en/latest/
+
+<a name="project-based"></a>
+
+## Project-based workflows
+
+In project-based workflows, a project is a directory containing a manifest
```
stefanv, line 56: So, who are we targeting here? I can think of two groups:

(1) User who got hold of some random package, now wants to get that package ready for use. Is it our responsibility to provide package development instructions? Probably not.

(2) A user who is starting out with their own application , and wants to add a mechanism for setting up the exact right environment, including SciPy. This is a more reasonable group to target—although again I'm not sure it's the right place. Maybe simply point to `uv` and `pixi` website documentation?

Either way, these users are advanced, so they won't be arriving at the SciPy docs, trying to figure out how to add SciPy to uv/pixi envs.

content/en/install.md
```diff
@@ -3,101 +3,149 @@ title: Installation
...
+<a name="system-package-managers"></a>
 
-# Install system-wide via a package manager
+## Installing system-wide via a system package manager
 
 System package managers can install the most common Python packages.
 They install packages for the entire computer, often use older versions,
-and don't have as many available versions.
+and don't have as many available versions. They are not the recommended
+installation method, but they are the simplest.
```
stefanv, line 88:
```suggestion
installation method.
```

Causes a lot of headaches, so wouldn't give any suggestion that this route is simple. E.g., on Debian you run into issues because you can't subsequently pip install without special flags.

content/en/install.md
```diff
@@ -3,101 +3,149 @@ title: Installation
...
+via the `scipy-stubs[scipy]` extra on PyPI, or the `scipy-typed`
+package on conda-forge.
+To get a specific version `x.y.z` of SciPy (such as `1.14.1`),
+you should install version `x.y.z.*` below:
+
+    python -m pip install "scipy-stubs[scipy]" # or
+    # versions for illustrative purposes
+    uv add "scipy-stubs[scipy]==1.14.1.*" # or
+    conda install "scipy-typed>=1.14" # or
+    pixi add "scipy-typed=1.15.0.*"
```
stefanv, line 148:
```suggestion
```

I suggest keeping this section short, to keep the overall document concise. These instructions seem fairly self-evident, and is likely also documented on the pypi / conda pages for `scipy-stubs`.

stefanv (commented):
Thank you @lucascolley!

Overarching comment: think of the target audience—who will be reading this guide, and why? Further keep things as simple and straightforward as possible, reducing choices only to those absolutely necessary.

content/en/install.md
lucascolley: > If you want to be more nuanced about it, you could say, for env based: "recommended for beginners", and for project-based: "recommended for advanced users".

see my comment below in the project-based section

content/en/install.md
lucascolley: right, I was trying to also implicitly accommodate `virtualenv` and whatever other venv-like tools are used. But I agree that just mentioning `venv` here more clearly is probably sufficient.

content/en/install.md
zanieb: I would disagree that people using `uv` / `pixi` are always advanced, we get _a lot_ of beginner questions, i.e., "this is my first time using a Python project".

Anyway, if the installation is ever more complicated than `uv add scipy` I'm happy to add a dedicated integration page to the uv documentation.

stefanv: My question is more: "what is the path that takes someone to the SciPy installation page to ask about uv" and then "did they land in the right spot, or should they be sent somewhere else". I.e., if they're asking the right question, relevant to SciPy, we should absolutely answer it. If there's a more general resource we can point to, best to do that (community wide).

lucascolley (commented):
thanks for the review Stefan!

content/en/install.md
stefanv: @lucascolley I think if you want to describe, from an absolute beginner's perspective, how to get an environment set up using uv, that's fine too. What was there felt like too high level for a newcomer to start with. You'd have to explain which directory they should be in, how to activate the environment, how often they need to do that, how to install packages (per your snippet above), etc. And then it really feels more like an alternative env manager, rather than "project based". People who want to set up projects *should* be able to find good guides out there (pyopensci?).

content/en/install.md
stefanv: It's pretty short, and it covers a couple of common use-cases, so I think OK to keep for now.

content/en/install.md
stefanv: Can we combine the code snippets, perhaps, so there's One Recommended Way of installing SciPy + stubs?

lucascolley: > One Recommended Way of installing SciPy + stubs?

On the one hand, most existing users will just want to now how to install the stubs on top of their existing SciPy installation. But on the other hand, installing both at once seems much more suitable for a "how to install SciPy" guide, and ensures versions stay in sync. I don't think I'm confident choosing one over the other.

content/en/install.md
lucascolley: > You'd have to explain which directory they should be in, how to activate the environment, how often they need to do that, how to install packages (per your snippet above), etc. And then it really feels more like an alternative env manager, rather than "project based".

Okay, sounds good. I guess that would mean we can avoid repeating similar instructions for `pip` and `conda`? If so, is the recommendation of `uv`/`pixi` for new users justified?

content/en/install.md
lucascolley: perhaps we just mention how to install both at once here, but also say something like "to install `scipy-stubs` and `scipy` at once". Then users with existing installations should be able to infer that they can just install `scipy-stubs` instead.

content/en/install.md
jorenham: isn't it reasonable to assume that `uv` and `pixie` users already know how to install `scipy` once they've seen that `pip install scipy` is an option?

content/en/install.md
jorenham: > I don't think the versioning of scipy-stubs is immediately self-evident without someone being told about it

Most stubs-only packages use a similar versioning scheme, but there are probably many that haven't used stubs-only packages before.

content/en/install.md
lucascolley: > isn't it reasonable to assume that uv and pixie users already know how to install scipy

Yeah, so I think `uv add scipy` in itself wasn't helpful. We should instead be targeting people who are not experienced with either `uv`/`pixi` or `pip`/`conda` yet.

content/en/install.md
```diff
@@ -3,101 +3,206 @@ title: Installation
...
+2.  Create a new project in a new subdirectory, by executing the following in a terminal:
 
-<a name="conda-install"></a>
+        uv init try-scipy
+        cd try-scipy
 
-# Installing via Conda
+    {{< admonition hint >}}
+The second command changes directory into the directory of your project.
+    {{< /admonition >}}
```
lucascolley, line 50: @jarrodmillman, I believe the linter is suggesting changes which would break the formatting.

lucascolley (commented):
I've rewritten the page, focusing on a beginner perspective. How does that look @stefanv ? And does the `uv` section look good to you @zanieb ?

content/en/install.md
```diff
@@ -3,101 +3,206 @@ title: Installation
...
+
+### Installing with `pip`
+
+1.  [Install Python](https://www.python.org/downloads/).
+
+2.  Create and activate a virtual environment with `venv`.
+
+   {{< admonition hint >}}
+See [the tutorial in the Python Packaging User Guide](https://packaging.python.org/en/latest/tutorials/installing-packages/#creating-virtual-environments).
+   {{< /admonition >}}
```
lucascolley, line 130: this indentation looks redundant - the admonition doesn't indent as there is no body of the bullet to bind to, it seems.

content/en/install.md
```diff
@@ -3,101 +3,206 @@ title: Installation
...
+[install-pixi]: https://pixi.sh/latest/
+
+2.  Create a new project in a new subdirectory:
+
+        pixi init try-scipy
+        cd try-scipy
+
+3.  Add SciPy to your project:
+
+        uv add scipy
```
tylerjereddy: Now I'm getting confused -- shouldn't `uv` be `pixi` here?

content/en/install.md
```diff
@@ -3,101 +3,206 @@ title: Installation
...
+## Environment-based workflows
+
+In environment-based workflows, you install packages into an environment, which you
+can activate and deactivate.
+These workflows are well-established,
+but lack some reproducibility benefits of project-based workflows.
+
+### Installing with `pip`
+
+1.  [Install Python](https://www.python.org/downloads/).
```
tylerjereddy, line 124: Of course, you can use `uv` to do this as well and then switch back to being a simple `venv` dinosaur like me. But there are indeed already some comments about choice paralysis I think.

content/en/install.md
tylerjereddy: Looks like you fixed this one just now.

content/en/install.md
```diff
@@ -3,101 +3,206 @@ title: Installation
...
-# Installing via Conda
+    {{< admonition hint >}}
+The second command changes directory into the directory of your project.
+    {{< /admonition >}}
 
-You can install SciPy from the `defaults` or `conda-forge` channels with `conda`:
+3.  Install Python:
 
-    conda install scipy
+        uv python install
```
zanieb, line 54: I'd omit this — by default, we'll install Python if it's not installed yet. I don't have a strong preference though.

zanieb: (Similar to the Pixi instructions)

content/en/install.md
```diff
@@ -3,101 +3,206 @@ title: Installation
...
-commonly used packages and tools. These downloadable files require
-little configuration, work on almost all setups, and provide all the
-commonly used scientific Python tools.
+## Project-based workflows
 
-[Anaconda](https://www.anaconda.com/download/) works on Windows, Mac,
-and Linux, provides over 1,500 Python packages, and is used by over 15
-million people. Anaconda is best suited to beginning users; it provides
-a large collection of libraries all in one.
+In project-based workflows, a project is a directory containing a manifest
```
stefanv, line 24: This is, according to the new text above, the recommended route for newcomers. So, the first thing a newcomer reads is "manifest", "lock-file", "dependencies". Yikes!

How about these more common use-cases:

- I found a snippet on StackOverflow and want to play with it in IPython or Jupyter.
- I am reading the Scientific Python lecture notes and want to try some examples.
- I am studying the SciPy API guide / other docs and want to try some code snippets I see.

When you make project-based the default installation path, the instructions that come with tutorials won't work. Nothing will work, as stated below, without quite a bit of future reading.

So I'll reiterate my earlier suggestion that you make this as simple as possible for someone who just wants to get going to see how the project works. And my strong feeling is that that route is pip or conda, and not starting a new project.

stefanv (commented):
I don't want to be too loud, so I'll let others weigh in. I think my viewpoint has been clearly described: put yourself in the shoes of a newcomer.

Thanks for reviewing these instructions, @lucascolley!

content/en/install.md
lucascolley: > So, the first thing a newcomer reads is "manifest", "lock-file", "dependencies". Yikes!

Agreed 😅 I think this is solvable by moving this below to the discussion of how environment-based workflows are different. It made more sense before I swapped the sections.

> When you make project-based the default installation path, the instructions that come with tutorials won't work. Nothing will work, as stated below, without quite a bit of further reading.

For my understanding, could you point to some tutorial instructions for which this would be problematic?

content/en/install.md
lucascolley: Yeah, I considered leaving out the hyperlink and letting users google "how to install Python" - could do that if people prefer. I don't think we should be laying out different options for installing Python.

content/en/install.md
lucascolley: omitted!

content/en/install.md
```diff
@@ -3,101 +3,206 @@ title: Installation
...
-- [WinPython](https://winpython.github.io): Another free distribution
-  including scientific packages and the Spyder IDE; Windows only, but
-  more actively maintained and supports the latest Python 3 versions.
-- [Pyzo](https://pyzo.org): A free distribution based on Anaconda
-  and the IEP interactive development environment; Supports Linux,
-  Windows, and Mac.
+[`uv`]: https://docs.astral.sh/uv/
 
-<a name="pip-install"></a>
+1. Install `uv`, following [the instructions in the `uv` documentation][install-uv].
```
lucascolley, line 35: @zanieb I wonder whether the instructions in the `uv` docs could be made even more beginner-friendly? Perhaps for people who aren't familiar with using a terminal, where you say "use `curl`", you could instead say something like "use `curl` in a terminal prompt", and maybe link out to an explanation of how to get a terminal/shell on different OSs?

lucascolley (commented):
thanks again @zanieb @stefanv ! I've tried to address the review comments.

> So I'll reiterate my earlier suggestion that you make this as simple as possible for someone who just wants to get going to see how SciPy works. And my strong feeling is that that route is pip or conda, and not starting a new project.

I agree that we should be making the first option as simple as possible - I just disagree that using pip or conda is any more simple than the instructions shown here for uv or pixi. While you can do a lot more with a project than a simple environment, I don't think that complexity seeps into the instructions here.

---

@rgommers does this look good to you after the substantial changes?

rgommers:
> I just disagree that using pip or conda is any more simple than the instructions shown here for uv or pixi.

I agree with `uv`/`pixi` actually being simpler and friendlier for beginners. What those beginners struggle the most with is the activation step of environments (whether venv or conda env), and `uv`/`pixi` removing the need to ever think about activation is a huge benefit.

rgommers:
For context, we always said "use a Python distribution" for beginners, because explaining activation and environment management to them was simply too challenging. But distributions are kinda dying, and we also posed some problems on the learning path for beginners by giving them a "base" environment to mess up. So I'd say lead with `uv`/`pixi` now, since it has become the best option for _both_ beginners and power users.

stefanv:
Sounds like y'all prefer the uv route, so in that case, can you think through the path a user would take who just wants to try a code snippet? For them, the notion of "project" may be confusing. They may or may not expect to have to activate an environment. How do they go from "I want to run a SciPy code snippet" to running that snippet, and once they've rebooted their computer, how do they do it again.

I think it is a very reasonable expectation for a new user that `python myscript.py` should Just Work (TM). They should not need to do `<some tool> run myscript.py`.

content/en/install.md
```diff
@@ -3,101 +3,206 @@ title: Installation
...
+You can also install SciPy and `scipy-stubs` as a single package,
+via the `scipy-stubs[scipy]` extra on PyPI, or the `scipy-typed`
+package on conda-forge.
+To get a specific version `x.y.z` of SciPy (such as `1.14.1`),
+you should install version `x.y.z.*`, for example:
+
+    uv add "scipy-stubs[scipy]==1.14.1.*" # or
+    pixi add "scipy-typed=1.15.0.*" # or
+    python -m pip install "scipy-stubs[scipy]" # or
+    conda install "scipy-typed>=1.14"
```
stefanv, line 205: This section reads very nicely now, btw, thanks!

In conda, does scipy-typed automatically match up with SciPy? If so, then this section is :ok_hand:

lucascolley: > In conda, does scipy-typed automatically match up with SciPy?

Yep! The relevant lines from the recipe are https://github.com/conda-forge/scipy-stubs-feedstock/blob/fc110f760877e2b8cfe169cb9f54d308ffeb7bd9/recipe/meta.yaml#L38-L47 and https://github.com/conda-forge/scipy-stubs-feedstock/blob/fc110f760877e2b8cfe169cb9f54d308ffeb7bd9/recipe/meta.yaml#L1-L2.

jorenham: FYI, `scipy-stubs==1.15.0.0` is [almost ready](https://github.com/jorenham/scipy-stubs/issues/250), so I'll definitely be able to release it *before* the `scipy==1.15.0` final release.

content/en/install.md
stefanv: It looks like the uv folks are urging people to [switch away from activating virtual environments](https://github.com/astral-sh/uv/issues/1910#issuecomment-2498847428). And they have good technical reasons for it.

That's all fine if you've decided to switch everything to UV, but it feels problematic from a higher level UI perspective. `python myscript.py` now becomes `some-tool-name run myscript.py`, so it (a) introduces a layer on top of Python (not ideal for newcomers) and (b) binds you into a specific, non-general invocation pattern.

As I mentioned before, I suspect this pattern might break a lot of existing tutorial advice, scripts with `#!/usr/bin/env python`, etc. In fact, that's a question I'm very curious about: how does `uv` intend for developers to use it if they require `uv run` for everything? (I'm speaking from complete ignorance here, of course, since I haven't studied uv at all.)

I also don't like that users now need to know about `uv` and `pip` and `uv pip`. It's often possible to plot some narrow trajectory with these tools where you hide much of the complexity (and we should have a community guide from that, perhaps via PyOpenSci, if they don't already). But then things break and you have a lot of concepts to understand before you can debug, and only uv-specific resources are helpful.

Anyway, these are just my high-level concerns; I'm sure they can all be addressed in one way for another, and I appreciate you having patience with me as I iterate them.

ev-br: Adding to what Stefan said --- and I also never used uv --- how does the workflow with a test script in a separate repository work with auto-activation?

```
$ mamba activate scipy-dev
$ python dev.py shell
$ cd ~/tmp/
$ python failing_script_with a breakpoint.py
```

Same question about working on a project (scipy) and its dependency (scipy-doctest):

```
$ mamba activate scipy-dev
$ cd scipy-doctest
$ hack hack hack
# ..... in a second terminal window
$ mamba activate scipy-dev
$ cd scipy
$ python dev.py test 
```

rgommers: > Same question about working on a project (scipy) and its dependency (scipy-doctest)

The one word answer is "workspaces" (implemented in `uv`, see https://docs.astral.sh/uv/concepts/projects/workspaces/, work in progress in `pixi`). In `pixi` right now you can do what you must also do with pip/conda: you manually install one then develop the other, without tracking the dependency between these packages. There is no good way to do it today outside of workspaces (and no, editable installs don't count since those are inherently incomplete).

ev-br: Uhm, it can't be that uncommon to have to fix a dependency? The behavior is
- in the environment I've an editable install of a dependency (scipy-doctest, but it can be numpy or anything)
- in one terminal I'm in the dependency tree, in the "bugfix" branch where I'm adding fixes
- in the other terminal I'm in the scipy tree, where I'm running scipy tests; might be also adding fixes to match

That it's an editable install is not very relevant (can be any other way to install the dependency _in this specific environment_), the choice of the env manager is irrelevant (can be mamba or venv or virtualenvwrapper or ...), the pair of projects can be anything. The main point is the ability to have two unrelated project / git trees in the same environment.

zanieb:
> I think it is a very reasonable expectation for a new user that python myscript.py should Just Work (TM). They should not need to do <some tool> run myscript.py.

While I agree with your sentiment, I don't know if 

```
uv init --script myscript.py
uv add --script myscript.py scipy
uv run myscript.py
``` 

is more complicated than

```
python -m venv .venv
source .venv/bin/activate
pip install scipy
touch myscript.py
python myscript.py
```

and in the later case you need to remember to activate the environment before executing the file or it will fail.

It'd be nice to have a PEP 723 aware `python` shim, but that's pretty controversial for non-beginners :)

content/en/install.md
stefanv: I think Python has to become uv run everywhere.

content/en/install.md
matthew-brett: This is related to the "projects" discussion, but Homebrew has recently made their Python installation considerably less friendly to beginners, by implementing the "virtualenv-only pip" PEP - and in particular, strongly deprecating `--user` installs - see discussion at https://github.com/orgs/Homebrew/discussions/3404 .

matthew-brett:
I must say I am also worried the "project" workflow as the default *for beginners*.

The problem with the project workflow is that the person using the workflow really does have to understand "environments", and this is very advanced for a beginner.

I say that as someone who has been teaching Python to beginners for a long time, to a lot of people.

A beginner may start off with a hazy understanding of a filesystem, let alone the structure of a Python environment - and I'm speaking from experience teaching Berkeley statistics students - and lots of other, less experienced students.

In a recent third-year undergraduate data science class, in a group very used to the Python data science stack and notebooks, we accidentally ran into `pyenv` and caused substantial and long-lasting confusion - because they had no model of what a Python environment was.  My appeal here is to consider the possibility that environments are more advanced and more difficult than we (experienced developers) might think.  And for that reason, I would personally only suggest using `pixi` or `uv` for experienced developers - for whom they are very useful tools.   We should surely point experienced developers to those tools and it would be perfectly reasonable to tell beginners - you should look into environments as you become more experienced, because they are useful in various ways.

For example, look at the second hit for "scipy tutorial" on Google just now:

https://www.w3schools.com/python/scipy/index.php

It has the following first example:

```
from scipy import constants

print(constants.liter) 
```

I'm sure we'd all agree that we want to get the beginner to that state, where they are concentrating on using Scipy, as quickly and easily as possible - and in particular, if possible - without needing to understand (I would argue) difficult concepts like Python environments.

For example - I just ran the Python.org GUI installer on Mac, following all the defaults.   I then open Terminal and:

```
# Once only, won't need to think about this again.
pip3 install scipy ipython
ipython
```

- at which point I can copy / paste the example above, and it works.

If I close the terminal, and open it again, I don't have to activate anything, or be in a particular directory, I can just do:

```bash
ipython
```

and start copy-pasting.   It seems to me this is substantially easier for a beginner, such as a science undergraduate, than the per-project (`pixi`) or virtualenv-based workflows - and will better match much of the information already out there.

As long as we say - look, you'll probably need this environment stuff later - we will not hold them back as they get 6 or 12 months into their scientific Python journey.

rgommers:
> For example - I just ran the Python.org GUI installer on Mac, following all the defaults. I then open Terminal and:

That's just terrible though, there's a very good reason that Python packaging folks are trying so hard to discourage that. We should definitely not change to this method, it's a recipe for later disaster (well, sooner rather than later actually). If you want "easiest" in this fashion for the _absolute beginner_, then I think you're arguing for not changing the current defaults, which is "download Anaconda distribution, never use any packaging tool directly". 

Or, possibly even better, link to https://jupyter.org/try-jupyter/lab/, open the notebook and type `import scipy` - zero install for the win. I have met undergrads who used Pyodide/JupyterLite/PyScript like this for a while, and never had to deal with any packaging as a result.

content/en/install.md
rgommers: > but it can be numpy or anything

You actually can't with editable installs of `numpy` and say `scipy` or `scikit-learn`, it won't build the package that depends on `numpy`.

> The main point is the ability to have two unrelated project / git trees in the same environment.

That is correct. This is about the same with both methods (ignoring workspaces). In the `pip`/`conda` case you have to do `cd pkg1 && pip install .` then `cd pkg2 && pip install .  # or similar`, manually keeping track of the two packages both having a dev version installed, then running your tests. If you iteratively work on `pkg1` and `pkg2`, you just have to manage the state in your head. You can do that with `pixi` or `uv` as well - it's maybe very slightly more clumsy.

> Uhm, it can't be that uncommon to have to fix a dependency?

In the grand scheme of things, it is uncommon (>95% of users only consume rather than develop packages to start with). It's harder for project-based tools, hence I think it wasn't their first priority. But the workspaces concept was basically copied from Cargo, and seems like a good answer conceptually.

rgommers:
> Starter - use some installation method where environments are not visible - such as a Python.org install

I'd much prefer to keep pointing to Anaconda then if there must be a "no environments nor project managers" level, it's way better than python.org installers.

> very rarely leads to any problems for the beginner

This is not true I think beyond the duration of perhaps a single initial course with a limited set of packages. It works okay for a while, but using `pip install` repeatedly inevitably degrades an environment over time. And once it breaks, it's a huge pain for the user to understand what is going on or to fix it.

matthew-brett:
> > Starter - use some installation method where environments are not visible - such as a Python.org install
> 
> I'd much prefer to keep pointing to Anaconda then if there must be a "no environments nor project managers" level, it's way better than python.org installers.

I'm proposing that Python.org installers are totally fine for beginner science undergraduates - maybe we should discuss that if you think that isn't true?  But really - any methods (except Anaconda - due the Terms of Service) that don't require the user to think about or interact with environments would be OK.   Unless I've misunderstood, I think that excludes `uv` and `pixi`.

> > very rarely leads to any problems for the beginner
> 
> This is not true I think beyond the duration of perhaps a single initial course with a limited set of packages. It works okay for a while, but using `pip install` repeatedly inevitably degrades an environment over time. And once it breaks, it's a huge pain for the user to understand what is going on or to fix it.

I can tell you that I have taught many students now, and problems like that are rare, and generally easy to fix with the proposed methods that come up on StackOverflow.   My students for example, start off with the instructions:

```
pip3 install scipy matplotlib pandas statsmodels scikit-learn seaborn ipython jupyter
```

and generally need very little else for a year or two.  By which time time they are in a much better position to adopt virtualenv workflows.

Do you think it isn't true that beginners will find environments like `pixi` confusing?  Would it help if I did a walkthrough to show where they are likely to get confused?

lucascolley:
> But really - any methods (except Anaconda - due the Terms of Service) that don't require the user to think about or interact with environments would be OK.

How about we just stick with JupyterLite then? Do we need a level between that and environments?

matthew-brett:
> > But really - any methods (except Anaconda - due the Terms of Service) that don't require the user to think about or interact with environments would be OK.
> 
> How about we just stick with JupyterLite then? Do we need a level between that and environments?

As I was saying above - it seems overly strict to insist that people stick with JupyterLite or go straight to understanding environments.  JupyterLite has some serious downsides, including the difficulty of installing new packages, and the fact that everythings sits in the browser cache.   So, given there are ways of installing that don't require understanding of environments, it seems reasonable to me to point beginners, specifically, at those, while saying they will in due course graduate to environments.

tupui:
> I'm proposing that Python.org installers are totally fine for beginner science undergraduates - maybe we should discuss that if you think that isn't true? But really - any methods (except Anaconda - due the Terms of Service) that don't require the user to think about or interact with environments would be OK. Unless I've misunderstood, I think that excludes uv and pixi.

If you are looking for a sane one line install/use. UV exactly proposes that. Just not even present `pip`, not even `uv pip` but the new CLI and workflow they propose. It's a radical approach but I am getting more and more convinced this is going to be the way sooner than later.

matthew-brett:
> > I'm proposing that Python.org installers are totally fine for beginner science undergraduates - maybe we should discuss that if you think that isn't true? But really - any methods (except Anaconda - due the Terms of Service) that don't require the user to think about or interact with environments would be OK. Unless I've misunderstood, I think that excludes uv and pixi.
> 
> If you are looking for a sane one line install/use. UV exactly proposes that. Just not even present `pip`, not even `uv pip` but the new CLI and workflow they propose. It's a radical approach but I am getting more and more convinced this is going to be the way sooner than later.

I'm totally happy to be convinced that `uv` can allow one to avoid understanding or using environments - in fact I've been in that discussion over at https://github.com/astral-sh/uv/issues/2077 - could you say more how one would do that?

tupui:
> I'm totally happy to be convinced that `uv` can allow one to avoid understanding or using environments - in fact I've been in that discussion over at [astral-sh/uv#2077](https://github.com/astral-sh/uv/issues/2077) - could you say more how one would do that?

Depends what you want to do. `uv run --with scipy example.py` is quite simple. Their guides are pretty straightforward even for beginner https://docs.astral.sh/uv/guides/ if you are used to dancing with envs this is actually harder to wrap your head around all that because its a different take and vision.

matthew-brett:
> > I'm totally happy to be convinced that `uv` can allow one to avoid understanding or using environments - in fact I've been in that discussion over at [astral-sh/uv#2077](https://github.com/astral-sh/uv/issues/2077) - could you say more how one would do that?
> 
> Depends what you want to do. `uv run --with scipy example.py` is quite simple. Their guides are pretty straightforward even for beginner https://docs.astral.sh/uv/guides/ if you are used to dancing with envs this is actually harder to wrap your head around all that because its a different take and vision.

The use case is the most basic one possible - how do I get up and running with a Python environment that includes Scipy.   For the moment, I have no need for a Python environment that doesn't include Scipy - or any other environment.  When I install a package, I want to be able to import it without further ado from my terminal IPython or Jupyter.

This is the model that most tutorial out there will use.

With Python.org I run my basic install command, e.g.:

```bash
pip3 install scipy ipython
```

and from then on, I can `import scipy`.  

However `uv run --with scipy example.py` prompts the following questions:

* What is this `uv run` command?  Why isn't the command `python3` as it is for most other tutorials?
*  Why do I have to specify all my dependencies on the command line there?   Why can't I just install `scipy` and be done with it?  I'm going to keep using Scipy for everything.

So - yes - sure - if you really need separable isolated environments - `uv` or `pipx`.  If not - which is the case for scientists who aren't experienced coders, and may well never be experienced coders - then give them the simplest possible thing, that will match what they see out in the interwebs.

tupui:
> What is this uv run command? Why isn't the command python3 as it is for most other tutorials?

That's my point. UV is too new for tutorials to be adapted to it. (But I always argue that many tutorial are either way not worth it vs the official docs of projects, but that's another discussion).

True newcomers won't care if you do `python3, python, pip, uv`. All is new for them.

But for real, I never understood why it's an issue to ask users to type two commands before running a script.

> Why do I have to specify all my dependencies on the command line there? Why can't I just install scipy and be done with it? I'm going to keep using Scipy for everything.

You can tell them to define the deps in the script. AFAIK, UV is still the only system which supports that cool PEP https://peps.python.org/pep-0722/.

matthew-brett:
I don't think you can assume beginners won't look around the web for other tutorials - in fact - I think you can assume they will.   So if we're on the bleeding edge, we're going to have users wondering whether scipy might be too advanced for them - for no good reason.

I'm sure you find this too - but most scientific programmers have collections of notebooks and scripts, all of which import the same libraries - `numpy`, `scipy`, `matplotlib` etc.   So, if we set them up with a system that provides something they don't need (isolation of environments) while requiring them to do something extra (specifying dependencies in scripts - (?and notebooks?) that seems to them burdensome and superfluous - that will also be harmful.

rgommers:
> I'm proposing that Python.org installers are totally fine for beginner science undergraduates - maybe we should discuss that if you think that isn't true? But really - any methods (except Anaconda - due the Terms of Service)

I'd like to reject this proposal and not re-litigate it (python.org installers + raw pip are still as bad as 5 or 10 years ago, nothing fundamentally changed here). The only change since the last time we made major updates to the install instructions is (a) the emergency of `uv` and `pixi`, and (b) the Anaconda Terms of Service changes - which I'm certainly not happy at all about, however they also do not affect educational institutions or beginning users (the material change is research institutes/nonprofits with >200 employees, xref [this post](https://www.anaconda.com/blog/update-on-anacondas-terms-of-service-for-academia-and-research)).

I think we either point to the new & improved tools, or change nothing.

matthew-brett:
Which proposal are you rejecting - all the changes in the PR here - or the proposal simply to relabel the pip way as the beginner way?    It would satisfy my concerns if the changes just reversed the order (from beginner to advanced) of the `uv`/`pixi` option with the `pip` / `conda` option.

For the Python.org installers - is this due to the fact that environments can get stale and inconsistent over time?   Is this really a veto for their use in simple cases - when that problem is rare?   At least, shouldn't we discuss / assess whether it is rare or not, rather than assuming it?   Our experience as developers will be very different from that of beginners. 

It seems to me this is a practicality over purity issue, where we should prioritize practicality for our beginner users, while acknowledging that more advanced users will want something that can protect themselves from those problems, partly because the cognitive load of using environments is less for those users, and partly because advanced users do much more complex and frequent installs.

rgommers:
> the proposal simply to relabel the pip way as the beginner way?

this

> is this due to the fact that environments can get stale and inconsistent over time? Is this really a veto for their use in simple cases - when that problem is rare?

it is absolutely not rare, it is quite common. there's a whole XKCD comic about it and uncountable issues, SO posts, etc.

I can't spend the time digging up the whole history here, but we had a very very long discussion across multiple projects about recommending Anaconda distribution vs. python.org+pip, which I'm sure you remember (Thomas Kluyver did a lot of the heavy lifting IIRC). Nothing really changed since then in this equation. 

> It would satisfy my concerns if the changes just reversed the order (from beginner to advanced) of the `uv`/`pixi` option with the `pip` / `conda` option.

If we don't want to recommend `uv`/`pixi` for beginners because a number of people here think that that is too hard and therefore not the right choice, then I want to stay with Anaconda as the recommendation.

zanieb:
@tupui just a heads up, pipx also supports running PEP 723 scripts, e.g. with`pipx run example.py` — it'd be nice if more tools supported _writing_ the inline metadata because it is sort of a pain for beginners. But... you could provide a more tool agnostic guide by having a blurb of line metadata someone can copy into a script?

matthew-brett:
> @tupui just a heads up, pipx also supports running PEP 723 scripts, e.g. with `pipx run example.py` — it'd be nice if more tools supported _writing_ the inline metadata because it is sort of a pain for beginners. But... you could provide a more tool agnostic guide by having a blurb of line metadata someone can copy into a script?


> > the proposal simply to relabel the pip way as the beginner way?
> 
> this
> 
> > is this due to the fact that environments can get stale and inconsistent over time? Is this really a veto for their use in simple cases - when that problem is rare?
> 
> it is absolutely not rare, it is quite common. there's a whole XKCD comic about it and uncountable issues, SO posts, etc.

I think it is rare - for beginners - because I've been there when lots of beginners have done lots of installs.    So, even if there is an XKCD comic on it (didn't that include `easy_install`?) I don't think that question is outside the range of reasonable discussion.

> I can't spend the time digging up the whole history here, but we had a very very long discussion across multiple projects about recommending Anaconda distribution vs. python.org+pip, which I'm sure you remember (Thomas Kluyver did a lot of the heavy lifting IIRC). Nothing really changed since then in this equation.
> 
> > It would satisfy my concerns if the changes just reversed the order (from beginner to advanced) of the `uv`/`pixi` option with the `pip` / `conda` option.
> 
> If we don't want to recommend `uv`/`pixi` for beginners because a number of people here think that that is too hard and therefore not the right choice, then I want to stay with Anaconda as the recommendation.

OK - but just to clarify - what, specifically, does Anaconda get you beyond the effect of:

```bash
pip3 install scipy matplotlib pandas statsmodels scikit-learn seaborn ipython jupyter
```

?  I think I'm saying that that issue is not as resolved as you think it is, and bears further discussion.

rgommers:
> what, specifically, does Anaconda get you beyond the effect of:

Off the top of my head:
- many other popular packages that beginning to intermediate users may want/need, with integration testing between them,
- a GUI for launching Python or Jupyter notebook/lab, and for navigating packages, with OS menu entries and other such niceties for beginners, and 
- not needing to use `pip` or `conda`, and hence not teaching the bad habits of installing into a base env or with `--user`.

>  I think I'm saying that that issue is not as resolved as you think it is, and bears further discussion.

It is resolved for SciPy, because you have nothing new to say and I am unwilling to rehash the old argument - I have more productive and less draining things to do. You have a deep, and in my opinion unjustified, dislike of anything that smells like it comes from a company, or perhaps Anaconda in particular (maybe it now extends to Astral and Prefix too, dunno). That's your good right, and you get to use and teach with the tools of your preference. I'm not really interested in that bias though, nor in re-discussing this _exact same issue_.

We can either consider new project based tools as the new recommended method to get started (where we started this discussion from), or stay with the Anaconda distribution, or some other new change like pointing at JupyterLite.

matthew-brett:
> > what, specifically, does Anaconda get you beyond the effect of:
> 
> Off the top of my head:
> 
>     * many other popular packages that beginning to intermediate users may want/need, with integration testing between them,
> 
>     * a GUI for launching Python or Jupyter notebook/lab, and for navigating packages, with OS menu entries and other such niceties for beginners, and
> 
>     * not needing to use `pip` or `conda`, and hence not teaching the bad habits of installing into a base env or with `--user`.

`jupyter lab` from the command line is an adequate replacement for the GUI features of Anaconda, and the packages that I listed before were certainly enough to teach all three years of our data science courses at my last university.   Of course, if one has to go further, one is no longer working with beginners.

> > I think I'm saying that that issue is not as resolved as you think it is, and bears further discussion.
> 
> It is resolved for SciPy, because you have nothing new to say and I am unwilling to rehash the old argument - I have more productive and less draining things to do. You have a deep, and in my opinion unjustified, dislike of anything that smells like it comes from a company, or perhaps Anaconda in particular (maybe it now extends to Astral and Prefix too, dunno). That's your good right, and you get to use and teach with the tools of your preference. I'm not really interested in that bias though, nor in re-discussing this _exact same issue_.

Ah - no - I think you're assuming that my opinions are based on anti-corporate bias, but that's not the case.  In fact, the company angle hadn't even occurred to me for `uv`.   So I wonder whether you'd reconsider dismissing my opinions on these matters - they really are based on a lot of experience working with beginners - over a long period - and with many iterations.   I suspect I may be unusual in this conversation in working with many beginners, but I'm happy to be corrected.

The situation has absolutely changed from the very old discussions we've had on this in the past - as you can see from looking at the XKCD cartoon you referred to:

https://xkcd.com/1987

And in fact, the cartoon is, if anything, complaining about the great excess of different methods of install, so adding `uv` and `pixi` would only add more loops and lines.

And although it may be tiring, it is an important discussion, because the wrong install instructions will have a serious impact on Scipy's popularity.

rgommers:
No, I am not willing to reconsider changing to python.org + pip. With all due respect, please consider it vetoed. That's a very rare thing for me to say, but I really cannot afford to spend mental energy arguing over this.

> I think you're assuming that my opinions are based on anti-corporate bias, but that's not the case

Yes, I am assuming that given your history on this topic (does not seem like an unfair assumption to me), but even if that's not the case, I am 100% certain that we are not going to agree on the technical merits anyway.

matthew-brett:
On anti-corporate bias - it's true that I was worried that the open-source community would commit itself too deeply to a product like Anaconda, where there was no guarantee anyone could sensibly offer that it would not be monetized.    Therefore I wasn't surprised - or upset - when Anaconda did do that with its initial change to the terms of service, that would have required reseachers at universities to pay a fee to install Anaconda packages.  And the situation would have been a lot worse if we had more completely committed to Anaconda as an install method.  So I would characterize the position I took at the time as suitable caution in the face of potential conflicts of interest, rather than anti-corporate bias.

matthew-brett:
> No, I am not willing to reconsider changing to python.org + pip. With all due respect, please consider it vetoed. That's a very rare thing for me to say, but I really cannot afford to spend mental energy arguing over this.

Maybe the trick then, would be for you to drop out of the discussion for now - we can see if we can come to a reasoned and agreed conclusion, and then you can come back when that is ready?

rgommers:
I don't have much more to add Matthew. I strongly recommend not to waste anyone's time pushing your point here.

stefanv:
I don't think we need to escalate to using a veto?—I mean, Matthew could just as easily then veto the uv approach, and then we're stuck. Nor do I find it helpful to suggest that he's wasting people's time, when he's emphasizing a point that also appeared earlier in this thread. I happen to find it convincing from my own teaching experience.

I also understand that we don't want to spend forever "re-litigating" the issue, so how can we proceed?

Anaconda is not a suitable default recommendation anymore, because we may expose our users to risk. No one reads the installer license, so we do carry some responsibility in what we point our users to.

I feel like uv, as great a tool as it is, is not currently viable, since it (`uv run`) is incompatible with the tutorials and advice currently very widespread online. If the standard community workflow changes over time, and uv becomes the standard way of using Python (would be a bit bizarre to invoke Python as `uv`, but OK), then we should re-assess.

I am not as opposed as Matthew to teaching users how to activate a default environment, and I think they can be explained reasonably easily ("a virtual env is like the toolbox with all your Python tools in it; you need to open it up before doing any work"). And then the activation instructions are straightforward: 

- Install Python
- python -m venv ~/envs/py313
- Every time a terminal is opened: `source ~/envs/py313/bin/activate` (or pop into `~/.shellrc`) and you're good to go.

You're gonna laugh, but on my machine I have:

```
$ ls ~/envs
py311  py312  py313  py313t  py37  py38  py39
```

I mostly get away with one venv per Python :sweat_smile: :man_shrugging: 

Here are the guidelines I would suggest for SciPy installation instructions:

- User must be able to `pip install scipy` (or equivalent, such as `conda install scipy` or `uv pip install` etc.)
- User must be able to run `python myscript.py` (equivalently, a script with `#!/usr/bin/env python`)—and I mean that precise spelling, because that's what users are going to read in books.
- User must be able to change to a different directory and run `python myscript.py`

I think if you satisfy these conditions (and there may be a few others I missed), it would avoid too much confusion among new users.

That said, I will be away until January, so  won't be here to object if y'all decide to go in a different direction ;)

Happy holidays, everyone! :christmas_tree: I love that we get to build this project together, and appreciate that I get to learn and work with y'all.

ev-br:
My 2 cents from previous job, which involved teaching undergrads, too. For context: ~200-400 undergrads per year, from 1st to 3rd year applied maths, physics and "digital humanities", teaching numerical analysis classes; ~100 hallway ad hoc questions/suggestions from undergrads doing school projects using python/numpy/pydata stack libs; advising ~10 undergrads/MSc students per year doing semester/year long research projects.
The student levels were ranging from absolute greenhorn beginners (literally, first time using the keyboard not a touchscreen) to part-time developer job holders.

For beginners and teaching large classes: The ability have Anaconda installed was an absolute life-saver. Both having the university admins to install Anaconda in the university classes, and telling students "install Anaconda on your laptops", then showing once how to start Spyder (or what is the bundled IDE, accessible from the Windows Start menu) --- and then the setup is out of the picture.  
("I use a macbook and it does not have the Start button"--- "Install Anaconda and figure out how to start Spyder")

These days after the Anaconda TOC change, I'd push for the university getting the appropriate license from Anaconda.  

Going through the "pip3 install numpy scipy pandas matplitlib" route would have been _very_ disruptive. 

For students who are not absolute beginners or start doing projects beyond basic school curriculum, and outgrow the absolute beginner level, and start messing with the base environment, we of course observed all the issues Ralf mentioned above.

These problems are very real, and the typical solution was to hold a quick session on "hey, here are several things you now need to learn: here's the thing called environments, there's git,  and here's what you need to read up on". 
The "read up on" part was heavily relying on _pydagogue_ tutorials from Matthew (hey Matthew, huge thanks for these!--- the git one was a must-read, as were several others). 

So pragmatically, my suggestion for the scipy docs would be to
- keep it with Anaconda
- have a small piece of text explaining that there's TOC story + links to further reading
- link to instructions on how to set up conda-forge or what's the current way is (doesn't matter all that much IMO if it's conda-forge, mambaforge, miniconda, micromamba --- honestly, I'm not sure what's the current one)
- Mention that there are alternatives: uv, pixi, etc. Link to accessible tutorials if they are readily available.

If these explanations/tutorials  do not exist or are lacking, let's write more of these, place them somewhere and link from the scipy.org docs.

The _somewhere_ part is essential: the scipy docs are probably not the right place to follow emerging trends. Once those technologies stabilize and become boring mainstream, great, then we update the scipy docs---and update the links to then-current emerging tech.

matthew-brett:
Thanks @ev-br and @stefanv for your thoughtful comments.

It seems to me that the way forward here must be to move from assertion to argument, and if possible, to evidence.   I've put some thoughts below ("open questions").

Unfortunately, it's going to be very hard to get there if we are under threat of imminent veto - @rgommers - I guess you agree?   If so, I wonder whether we should move this important discussion to another place where there is more space for fruitful disagreement and debate?

## Open questions

* Is it reasonable to require beginners to open a terminal to run commands such as `pip install` or `jupyter lab`?  I think yes, and have experience to back that up - see for example https://textbook.nipraxis.org/installation_on_mac.html - but I'm happy for counterarguments.
* Is it reasonable to require beginners to edit their terminal configuration files?  I'd say no - in my experience this rapidly turns into a nightmare, especially on Windows.  I can firm that up with specifics if that's controversial.
* How often - in practice - do beginners find themselves with broken installations from some package incompatibility?   How hard - in practice - is it for beginners to get themselves out of this situation?   Can we make the escape route reasonably obvious for these situations? We should probably track down or recall actual examples for this - from StackOverflow or personal experience.
* How much does it matter that we give beginners commands that they don't understand - "magic"?   I'd say it does matter, and it matters more at the start than the middle of a course - the best students will always ask "why", and discomfort and anxiety usually follow.  I can give examples.
* How much do the current Anaconda terms of service matter?  Should we continue to worry that they will change again?
* If we do decide we need a one-click installer, how hard would it be to build our own, backed up by suitable CI, and that can be configured to install particular sets of packages?   I did this a long time ago for Mac : https://fullmonty.github.io - I guess it wouldn't be too hard to do that for Windows.

lucascolley:
I get the feeling that it is going to be difficult and time-consuming to reach agreement on all of those open questions. Could we perhaps instead focus on agreeing on some sort of improvement which we can merge, then continue further discussion elsewhere? The current instructions really are quite outdated.

Could those with concerns about the current diff in this PR distinguish between absolute blockers and things to consider going forwards? Concerns about Anaconda TOS and a standalone installer can't be blockers since currently Anaconda is recommended without any note about TOS, and such a standalone installer doesn't exist.

Perhaps we could even create a separate "absolute beginner" page which is targeted at users coming from their first ever tutorial or class, but which goes on to explain why those methods are not recommended to the interested reader?

matthew-brett:
> I get the feeling that it is going to be difficult and time-consuming to reach agreement on all of those open questions. 

With some effort and goodwill - maybe a week to get something that has been adequately discussed and agreed?   Although, as I said before, that might be harder here than elsewhere.   Still, if we do move it elsewhere, and then come back with the result, is a week an acceptable timeframe?

> Could we perhaps instead focus on agreeing on some sort of improvement which we can merge, then continue further discussion elsewhere? The current instructions really are quite outdated.

Honestly, if we need something now now, then I can live with the Anaconda option - but I do think it needs at least a mention and a link to the terms of service.   It seems to me it would be very easy for someone in a medium-sized company to install Anaconda without realizing they could become liable to the download fees, and it would be irresponsible of us to make that easier, given the ease with which we can help people avoid it.

lucascolley:
> With some effort and goodwill - maybe a week to get something that has been adequately discussed and agreed? Although, as I said before, that might be harder here than elsewhere. Still, if we do move it elsewhere, and then come back with the result, is a week an acceptable timeframe?

Would you be willing to open a PR yourself at that stage? I would prefer to limit the scope of my PR to the main changes I feel are important (mentioning static type stubs, mentioning `venv`s for pip, mentioning just Miniforge for conda installs, mentioning project-based workflows for users who are not absolute beginners, mentioning Anaconda's blog post about TOS if we keep Anaconda).

> Honestly, if we need something now now, then I can live with the Anaconda option - but I do think it needs at least a mention and a link to the terms of service. It seems to me it would be very easy for someone in a medium-sized company to install Anaconda without realizing they could become liable to the download fees, and it would be irresponsible of us to make that easier, given the ease with which we can help people avoid it.

Yes, that was what I originally had in this PR.

matthew-brett:
> Would you be willing to open a PR yourself at that stage? I would prefer to limit the scope of my PR to the main changes I feel are important (mentioning static type stubs, mentioning `venv`s for pip, mentioning just Miniforge for conda installs, mentioning project-based workflows for users who are not absolute beginners, mentioning Anaconda's blog post about TOS if we keep Anaconda).

That sounds reasonable to me.

> Yes, that was what I originally had in this PR.

Would you consider restoring that?   I think I must have missed the discussion.

lucascolley (commented):
The first thing you see now on the install page is the following tip:

image

If you click through to the beginner guide, you get three options. Firstly, a link out to https://jupyter.org/try-jupyter/lab/. Secondly, a slightly more concise version of the existing "Scientific Python Distributions" section, with links to Anaconda, WinPython and Pyzo. This includes the note:

> Users in large, non-university institutions may want to read Anaconda’s helpful blog on [“when is Anaconda free to use?”](https://www.anaconda.com/blog/update-on-anacondas-terms-of-service-for-academia-and-research)

Thirdly, `python -m pip install scipy`. However, this is accompanied with the warning:

> You may see this recommended in tutorials or classes, but it is not recommended to install packages globally on your system. The recommended way to install SciPy with pip is to use a virtual environment - see [Installing with pip](https://deploy-preview-595--scipy-org.netlify.app/install/#installing-with-pip).

and a link to the PyPA discussion of venvs for the interested reader.

---

For users who remain on the main installation guide, they are recommended the project-based workflows, with full instructions to get set up from scratch, and also given the option of environment-based workflows. The existing sections on system package managers and building from source are reproduced below. Then we get the section on installing with type stubs.

---

Please let me know if anyone sees any more blockers. I have also addressed @stefanv's concern regarding "what does a `uv` user need to do if they reboot their machine" in the `uv` section.

content/en/beginner-install.md
```diff
@@ -0,0 +1,75 @@
...
+To try out SciPy, you don't even need to install it!
+You can use SciPy in your browser at https://jupyter.org/try-jupyter/lab/ -
+just write `import scipy` in one of the notebook "cells" and hit play.
+
+For next steps, see [the SciPy user guide][scipy-user-guide].
+
+[scipy-user-guide]: https://docs.scipy.org/doc/scipy/tutorial/
+
+<a name="distributions"></a>
+
```
matthew-brett, line 30: Do they need any warnings here about losing their work if they clear their cache etc?

content/en/beginner-install.md
```diff
@@ -0,0 +1,75 @@
...
+commonly used packages and tools. These downloadable files require
+little configuration, work on almost all setups, and provide all the
+most commonly used scientific Python tools.
+[Anaconda](https://www.anaconda.com/download/) works on Windows, Mac,
+and Linux, and is best suited to beginning users.
+Other options include:
+
+- [WinPython](https://winpython.github.io): Another free distribution
+  including scientific packages and the Spyder IDE; Windows only, but
+  more actively maintained and supports the latest Python 3 versions.
```
matthew-brett, line 43: More actively maintained than Anaconda?

content/en/beginner-install.md
```diff
@@ -0,0 +1,75 @@
...
+- [Pyzo](https://pyzo.org): A free distribution based on Anaconda
+  and the IEP interactive development environment; Supports Linux,
+  Windows, and Mac.
+
+{{< admonition note >}}
+Users in large, non-university institutions may want to read Anaconda's
```
matthew-brett, line 49: How about"
```suggestion
Anaconda is free to use for inviduals, universities, and companies smaller than 200 employees.   For more detail, see Anaconda's
```

content/en/beginner-install.md
```diff
@@ -0,0 +1,75 @@
...
+{{< /admonition >}}
+
+After installing a scientific Python distribution,
+see next steps in [the SciPy user guide][scipy-user-guide].
+
+<a name="pip-global"></a>
+
+## Installing globally with `pip`
+
+If you already have Python installed, you can install SciPy globally
```
matthew-brett, line 60: Some explanation of "globally"?

matthew-brett (commented):
Some quick thoughts.

content/en/beginner-install.md
lucascolley: that sounds good to me

content/en/beginner-install.md
lucascolley: not sure - suggestions welcome!

content/en/install.md
```diff
@@ -3,101 +3,222 @@ title: Installation
...
+
+    conda install scipy # or
+    mamba install scipy
+
+[Miniforge]: https://conda-forge.org/download/
+[`mamba`]: https://mamba.readthedocs.io/en/latest/
+
+<a name="system-package-managers"></a>
+
+## Installing system-wide via a system package manager
```
stefanv, line 168: Since these methods are so problematic, we could consider only keeping the first sentence (warning) and not giving individual instructions for Debian, Fedora, and macos here.

Optionally, we could then show the route I mentioned elsewhere, which works perfectly well: install Python via system and then just create a virtual env using `python -m venv`. Since this page is no longer the beginner guide, I think that should be acceptable?

The suggestion is not needed for this PR, but if it seems like a sensible suggestion I'm happy to make a follow-up PR in the future.

stefanv (approved):
Thanks, @lucascolley, for addressing the concerns around a beginner workflow. With those changes, including Matthew's phrasing around Anaconda licensing, I'm +1 on merging.

content/en/install.md
```diff
@@ -3,101 +3,222 @@ title: Installation
...
 
 A word of warning: building SciPy from source can be a nontrivial exercise. We
-recommend using binaries instead if those are available for your platform.
+recommend using binaries instead if those are available for your platform
+via one of the above methods.
 For details on how to build from source, see
-[this guide in the SciPy docs](https://scipy.github.io/devdocs/building/index.html).
+[the building from source guide in the SciPy docs][building-docs].
+
+[building-docs]: https://scipy.github.io/devdocs/building/index.html
```
stefanv, line 204: Given the default recommendation to use `uv` for projects, I was wondering whether, in the future, we should adjust the above page accordingly?

I presume it can be used the same way pip is for SciPy development, adding a third code tab with `uv run python dev.py` and `uv pip` for installing dependencies?

lucascolley: > Given the default recommendation to use uv for projects, I was wondering whether, in the future, we should adjust the above page accordingly?

Yep! Although for SciPy, given the compiler dependencies, Pixi will probably be our best bet. I already use https://github.com/rgommers/pixi-dev-scipystack for all of my SciPy development. Windows support for PyTorch and JAX on conda-forge is still maturing (though actively - https://github.com/conda-forge/pytorch-cpu-feedstock/pull/231), but we may be able to use PyPI on Windows to fill some gaps if anyone needs it.

A third tab for `uv` is probably not so useful, since it's relatively obvious what to do by following along with the `pip` instructions, and you have to do the same setup of system deps.

> I presume it can be used the same way pip is for SciPy development, adding a third code tab with uv run python dev.py and uv pip for installing dependencies?

All that to say that we could do something like this. Or we could make a new page instead of adding a new tab, since there is much less to explain when most of the work is already done in writing the Pixi manifest. We may want to wait until somebody wants to try to merge Pixi support into the main repo, though.

content/en/install.md
lucascolley: > Since these methods are so problematic, we could consider only keeping the first sentence (warning) and not giving individual instructions for Debian, Fedora, and macos here.

I would be happy with that, let's see what others think.

> Optionally, we could then show the route I mentioned elsewhere, which works perfectly well: install Python via system and then just create a virtual env using python -m venv. Since this page is no longer the beginner guide, I think that should be acceptable?

How does that differ from https://deploy-preview-595--scipy-org.netlify.app/install/#installing-with-pip? Just installing Python with a system package manager rather than python.org? I said in another thread that I would be happy to remove the hyperlink to python.org, but I thought that listing out options for installing Python might be overkill.

> The suggestion is not needed for this PR, but if it seems like a sensible suggestion I'm happy to make a follow-up PR in the future.

👍

lucascolley:
@jarrodmillman just pinging again to ask about the lint errors - do you know if there is a way to format indented admonitions in the theme which appeases the linter?

content/en/install.md
stefanv: Thanks, Lucas!

content/en/beginner-install.md
lucascolley: @CAM-Gerlach thought I'd ping in case you have any opinions on the Python Distributions section here! Hasn't been updated in a while...

content/en/beginner-install.md
matthew-brett: How about just "globally (for all users on the system)"?

content/en/beginner-install.md
```diff
@@ -0,0 +1,75 @@
...
+- [Scientific Python Distributions](#distributions)
+- [Installing globally with `pip`](#pip-global)
+
+<a name="jupyterlite"></a>
+
+## JupyterLite
+
+To try out SciPy, you don't even need to install it!
+You can use SciPy in your browser at https://jupyter.org/try-jupyter/lab/ -
+just write `import scipy` in one of the notebook "cells" and hit play.
```
lucascolley, line 23: to be explicit, should mention opening a Python notebook here

lucascolley:
any further comments, on the unresolved comments or otherwise? If not, let's merge this in a few days

content/en/install.md
jarrodmillman: You could try something like
```
<!-- prettier-ignore-start -->
...
<!-- prettier-ignore-end -->
```

stefanv (approved):

content/en/beginner-install.md
```diff
@@ -0,0 +1,78 @@
...
+
+After installing a scientific Python distribution,
+see next steps in [the SciPy user guide][scipy-user-guide].
+
+<a name="pip-global"></a>
+
+## Installing globally with `pip`
+
+If you already have Python installed, you can install SciPy globally
+(for all users on the system)
```
rgommers, line 63: I'd remove "globally (for all users on the system)" since that is fairly unlikely to be true. `python` is more likely to be installed in a per-user location, and if it's really global it may require `sudo`.

rgommers (approved):
This seems pretty good to merge, thanks Lucas and all.

content/en/beginner-install.md
stefanv: Makes sense 👍

stefanv:
@lucascolley the paragraph following is also impacted by that last change

lucascolley:
thanks Stefan, addressed

stefanv:
Thanks, Lucas! Sorry this got a bit more involved than anticipated ☺️

lucascolley:
> Sorry this got a bit more involved than anticipated ☺️

I wouldn't want it to be boring :) thank you to all reviewers who gave their time here!

jorenham:
Thanks Lucas!

PR #629, opened 2025-04-25
Add text for teams

goanpeca:
Adds a descriptive text for each of the teams :) 

Screenshot 2025-04-25 at 11-10-13 SciPy - Teams


tupui (approved):
Thanks 👍 that's the bare minimum we can say and accurate so approving :)

PR #648, opened 2025-05-20
Update image shortcodes

goanpeca:
Hello team!

As part of figure, image and button shortcodes as being updated on [this PR](https://github.com/scientific-python/scientific-python-hugo-theme/pull/674) I created this PR to prepare the website for moving to the new syntax when a new version is released!

---

Pinging @agriyakhetarpal

### DIFF
--- a/content/en/teams/emeritus-maintainers.toml
+++ b/content/en/teams/emeritus-maintainers.toml
@@ -1,179 +1,179 @@
 [[item]]
 type = 'card'
 classcard = 'text-center'
-body = '''{{< image >}}
-src = 'https://avatars.githubusercontent.com/u/136867?v=4"'
-alt = 'Avatar of Anne Archibald'
-{{< /image >}}
+body = '''{{< image
+  src= 'https://avatars.githubusercontent.com/u/136867?v=4"'
+  alt= 'Avatar of Anne Archibald'
+>}}
 Anne Archibald'''
 link = 'https://github.com/aarchiba'
 
 [[item]]
 type = 'card'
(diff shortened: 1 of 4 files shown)


agriyakhetarpal:
Hi @goanpeca, thank you! We'll also need to ensure that all of the new shortcodes use double quotes for the values of the parameters, as single quotes are not allowed. I'm not sure how this will work with cases where we need complex multi-line content – hopefully there aren't any.

goanpeca:
>Hi @goanpeca, thank you! We'll also need to ensure that all of the new shortcodes use double quotes for the values of the parameters, as single quotes are not allowed. I'm not sure how this will work with cases where we need complex multi-line content – hopefully there aren't any.

Will make the fixes!

goanpeca:
Made the fixes @agriyakhetarpal !

agriyakhetarpal (commented):
Thank you, @goanpeca! Perhaps it makes sense to update the figure, button, and image shortcodes in a consolidated PR, as one PR is better and would be easier to review than three (as I commented in https://github.com/pybamm-team/pybamm.org#283). The changes for the image shortcodes look good, though! I haven't run the website locally right now but we should confirm all pages before merging, of course.

goanpeca:
@agriyakhetarpal PRs will fix all. Some projects only use some of the shortcodes.

This one has been tested :)

agriyakhetarpal (approved):
Sounds good to me! Out of curiosity, what are the extra `\` and `""` (double double quote markers) at the end of the image URLs for?

